PR #48, opened 2023-12-13
Feature/rename dummyamr

genevievestarke:
This is to rename dummy arm wind modules to standin amr wind.
This resulted in changes to the standin amr wind module, the generate standin data module, and some bash and runscript edits.  There were also a few folder name changes.


misi9170 (approved):
Thanks @genevievestarke! This looks good---I checked out your branch and confirmed that the 02 example runs fine for me.

I found a few extra references to the dummy that I've switched and pushed up. These were all in comments/printouts, so don't affect the running code at all. I also added `logstandin` to the .gitignore and removed the file logstandin that had been added to the 02 example folder. If you wanted that in the repository, please feel free to reinstate it. 

I decided to leave the `logdummy` line in the .gitignore in case people still have those floating around after this change so they don't accidentally get added to the repository; and I also left references to the dummy in install_old.md, which I figure we'll soon be deleting anyway (probably when v1 is released). Those are the only two places (.gitignore and install_old.md) where the "dummy" naming is still referred to, I think.

Issue #35, opened 2023-11-29
Improve solar PV module


brookeslawski:
At the moment, the solar module computes power output solely based on a constant value of irradiance. It would be good to create an improved solar module that predicts power based on time-varying conditions such as irradiance, weather conditions, array type (i.e. fixed vs tracking), etc.

PR #58, opened 2023-12-14
Improved PV solar model

brookeslawski:
Improved the PV solar performance model in response to Issue #35 

- Added an example that predicts solar power of a generic 100 MW single-axis tracking PV plant and based on timeseries weather data.
- Added `solar_pysam.py`, which integrates NREL's `PySAM pvsam` to predict PV power.
- Added solar system design input file `100MW_1axis_pvsamv1.json`
- Added sample weather input file `NonAnnualSimulation-sample_data-interpolated-daytime.csv`, which was interpolated down to a time resolution of 0.5s from 1 min to match amrwind time-step

The installation instructions and the requirements will need to be updated to include PySAM v4.2.0 and restricting the Python version to 3.6.x-3.11.x.


docs/install_old.md
```diff
@@ -6,7 +6,7 @@
 # Installation
 Create a new conda environment for hercules:
 ```
-conda create --name hercules python
+conda create --name hercules python=3.11
```
brookeslawski, line 9: I'm not sure if the python version should be specified here or if there is a way to say here that any version 3.6-3.11 is okay. I know it's specified in `setup.py`, but to my knowledge, that file isn't used at this point in the installation.

docs/install_old.md
```diff
@@ -86,7 +96,7 @@ activate hercules`).
 
 In the first terminal, run
 ```
-helics_broker -t zmq  -f 2 --loglevel="debug"
+helics_broker -f 2 --consoleloglevel=trace --loglevel=debug --local_port=$HELICS_PORT &
```
brookeslawski, line 99: The original code doesn't work for me. It throws an error (shown below).
image

docs/install_on_kestrel.md
brookeslawski, line 1: I added a note in line 211, but these instructions are untested. I have only run hercules locally at this point.

docs/install_old.md
misi9170: This is a good point. I think this is OK; in many circumstances, the user will be adding to a larger environment and they'll skip this step, but for someone creating a standalone environment this works

environment.yml
```diff
@@ -0,0 +1,95 @@
...
+      - referencing==0.31.1
+      - requests==2.31.0
+      - rpds-py==0.13.2
+      - ruamel-yaml==0.18.5
+      - ruamel-yaml-clib==0.2.8
+      - scipy==1.11.4
+      - strip-hints==0.1.10
+      - tzdata==2023.3
+      - urllib3==2.1.0
+prefix: /Users/bstanisl/anaconda3/envs/hercules-pysam
```
misi9170, line 95: Is this file supposed to be included in the hercules repo?

misi9170 (approved):
All this looks good to me, and I've pulled and run the examples. Just double check whether environment.yml was intended to be included?

genevievestarke (approved):
Looks good!

PR #59, opened 2023-12-19
Add WHOC installation instructions to Hercules documentation

genevievestarke:


genevievestarke:
@misi9170 Should we have them switch to develop in the installation instructions or stay with main after it's released for Q1?

misi9170:
> @misi9170 Should we have them switch to develop in the installation instructions or stay with main after it's released for Q1?

I don't think we need to specify switching to develop is a necessary step. It could be worth putting a line in just mentioning that users wanting newer features should consider switching to develop?

genevievestarke:
Will do, that's added as a note to the documentation!

misi9170 (changes requested):

Issue #42, opened 2023-12-08
Time discrepancy between emulator and AMRwind


misi9170:
It appears that `absolute_helics_time` proceeds twice as fast as `AMRwind_time` (as though AMR wind is only run for half of the emulators time steps). 

For instance, in a loghercules log file, the AMR wind signals are always repeated. This can also be seen in the main_dict log csv files.

Is this intended, or is there some syncing issue between AMR wind and the main Hercules emulator?

PR #74, opened 2024-02-06
Fix sync issue

genevievestarke:
Fixing time syncing bug in emulator and amr standin. Addresses #42


misi9170 (approved):
Thanks for tackling this!

paulf81:
Yes, thank you @dvaidhyn and @genevievestarke !

PR #76, opened 2024-02-08
Enable passing power setpoints to wind simulators

misi9170:
[Floris v4](https://github.com/NREL/floris/tree/v4) allows wind turbines to be derated by passing them power setpoints. This pull request takes advantage of that functionality and implements it in the `FlorisStandin`. However, AMRWind does not currently handle derating turbines, so this code will not change the power output from AMRWind. Additionally, we will need to check that when `power_setpoints` are passed to AMRWind, they are simply ignored, rather than throwing an error.

I am currently working on a pull request for WHOC that will take advantage of this derating capability in Hercules (with the `FlorisStandin`). Once that PR is in place and I'm happy that both features work, I'll change the DRAFT status of this PR.


hercules/amr_wind_standin.py
```diff
@@ -160,18 +160,24 @@ def run(self):
...
-            try:
-                message_from_server = list(ast.literal_eval(incoming_messages))
-            except Exception:
+        for _ in range(1):
+            incoming_messages = self.helics_connector.get_all_waiting_messages()
+            if incoming_messages != {}:
+                try:
+                    message_from_server = list(ast.literal_eval(incoming_messages))
+                except Exception:
+                    print("here here")
```
genevievestarke, line 169: Some print statements here

genevievestarke (approved):
Looks good!

Issue #70, opened 2024-02-06
Create Floris to electrolyzer example


genevievestarke:
Combine examples 08 and 05 to create a Floris to electrolyzer example for electrolyzer control development

PR #77, opened 2024-02-11
Feature/floris electrolyzer example

genevievestarke:
This pull request replaces the AMR wind standin in Example 05 with the Floris standin.  This will be updated when Floris is able to take in power set points.


misi9170:
Will address #70

example_case_folders/05_floris_wind_standin_and_electrolyzer/bash_script.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Set the helics port to use: 
+export HELICS_PORT=32000
+
+#make sure you use the same port number in the amr_input.inp and hercules_input_000.yaml files. 
+# Set up the helics broker
+
+helics_broker -t zmq  -f 2 --loglevel="debug" --local_port=$HELICS_PORT & 
+python3 hercules_runscript_CLcontrol.py hercules_input_000.yaml >> loghercules_cl 2>&1 &
+python3 floris_runscript.py amr_input.inp amr_standin_data-higher.csv >> logfloris_cl 2>&1
```
misi9170, line 18: I think that the `amr_standin_data-higher.csv` argument is not needed here? Or perhaps it is, but it doesn't appear in the repository, so I wasn't able to run the example with this line as is. However, it does run when I take out the `amr_standin_data-higher.csv` argument though (but I'm not sure if it's running the right thing... :) )

genevievestarke: Yep, I changed the name for this and didn't update it. It should be floris_standin_data.csv and is updated now!

misi9170: Thanks! While you're at it, do the output files need the `_cl` appended? This causes them to be missed by the .gitignore, so it might be best to name them `loghercules` and `logfloris`---or, update the .gitignore to ignore `loghercules*` and `logfloris*`, which might be a better general solution?

misi9170: The suggested updates to .gitignore are now in #93

PR #84, opened 2024-02-27
Feature/update example 6

paulf81:
This pull request updates the example 06_amr_wind_standin_and_battery, as well as other surrounding code as a test run to make a number of changes to documentation, continuous integration and outputs.  Specifically:

- The batch script is made clearer, and runs the prerequisitie data generation script
- All outputs are sent to an output folder, which is deleted and recreated at the start of the run
  - This change is implemented in a few places within the main hercules code, and some checks that the outputs/ folder exists are added
- The README script is standardized into a form that can be brought to all examples
- The docs are reorganized to import the README files from the individual examples
- A new CI script is added to test run the examples (for now just example 06) to provide further test coverage
- Ruff linting and formatting turned up a few small changes


genevievestarke (approved):
Looks good to me!
Just wondering why we added back in netcdf here, as a requirement?

PR #85, opened 2024-02-27
[BUGFIX] Reenable AMR-Wind by setting up empty power_setpoints

misi9170:
#76 Enabled passing power_setpoints to the `FlorisStandin`; however, @jfrederik-nrel noticed that this created a bug when running with AMR-Wind, because the helics connection in AMR-Wind was not expecting extra elements in the `"control"` message being passed to it.

This PR fixes that bug by handling variable length control signals (without `power_setpoints` for AMR-Wind; and (optionally) with `power_setpoints` for the `FlorisStandin`).

### DIFF
--- a/hercules/amr_wind_standin.py
+++ b/hercules/amr_wind_standin.py
@@ -189,8 +189,16 @@ class AMRWindStandin(FederateAgent):
 
             # Compute the turbine power using a simple formula
             if self.message_from_server is not None:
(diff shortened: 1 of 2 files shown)


jfrederik-nrel (approved):
Confirmed that it handles turbine power from AMR-Wind properly now

misi9170:
> Confirmed that it handles turbine power from AMR-Wind properly now

Thanks!

Issue #73, opened 2024-02-06
Add citation button to github repository


genevievestarke:

misi9170:
Addressed by [ca4fcda](https://github.com/NREL/hercules/commit/ca4fcda0bc1fe7754bc8ddd4c57be0bb3ee10c39), based on work by @ZackTully in #86.

PR #86, opened 2024-02-29
Citation file

ZackTully:
#73

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,10 @@
+# https://citation-file-format.github.io/about/
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
+authors:
+  - family-names: NREL
+    given-names: 
+title: "HERCULES"
+version: 1.0
+
+date-released: 2023-12-20
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -6,5 +6,6 @@ authors:
     given-names: 
 title: "HERCULES"
 version: 1.0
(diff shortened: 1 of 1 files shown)


misi9170:
Just noting here that I think the citation button won't go into affect until develop is merged into main. An option would be to point this PR at NREL:main instead, which should not create any complications with later merging develop into main (or if it does, they will be easy to manage). @genevievestarke any opinions on that?

genevievestarke:
Sounds good to me!  I think it might be easier to push this file to main, and then merge back into develop?

genevievestarke (approved):

misi9170:
> Sounds good to me! I think it might be easier to push this file to main, and then merge back into develop?

Ok; I think to do that, the best way will be for me to check out this file on main; push directly to main; merge main into develop; and push develop. This PR will remain open during that process, but once that is done, I can "merge" this PR to show it's completion (although, at that point, there should be no difference between develop and this PR).

misi9170:
I've now directly pushed CITATION.cff to main [ca4fcda](https://github.com/NREL/hercules/commit/ca4fcda0bc1fe7754bc8ddd4c57be0bb3ee10c39), and then merged main back into develop [f598a1a](https://github.com/NREL/hercules/commit/f598a1ac916bcef417d10c98fb0618dcf6d59968). To avoid mucking up the history, I'm closing this PR _without_ merging, and will then close #86.

PR #88, opened 2024-03-01
Added example with AMR-Wind + OpenFAST

jfrederik-nrel:
In "example_case_folders/09_amr_wind_openfast", a new example has been added that shows how to run Hercules with an OpenFAST turbine in AMR-Wind.


genevievestarke (commented):
Does this have a bash script associated with it? It seems to be missing an amr inp file and a bash script.

PR #89, opened 2024-03-11
Added AMR-Wind+OpenFAST example (including batch script)

jfrederik-nrel:
Sorry @genevievestarke, I accidentally removed the previous pull request as I'm clearly still figuring out this whole Github thing. This pull request should have all the files you need. Let me know if you're still missing something.


hercules/emulator.py
```diff
@@ -396,7 +396,7 @@ def process_periodic_publication(self):
         tmp = np.array(
             [self.absolute_helics_time, self.wind_speed, self.wind_direction]
             + yaw_angles
-            + power_setpoints
+            #+ power_setpoints
```
genevievestarke, line 399: Does amr wind not run unless the power_setpoints are commented out?

genevievestarke:
Things that need to change for each user:

1. batch script (location of AMR wind and spack environment
2. libdiscon.so file in ServoDyn.dat file

hercules/emulator.py
misi9170: Yeah ok so #85 avoids the need to comment out the `power_setpoints` by making `power_setpoints` an empty list (rather than a list of `None`s) if running with AMR-Wind. This line should be reverted (uncommented), but then you'll need to merge develop back in to get the changes from #85 in order to run.

genevievestarke:
Should we run Hercules at a smaller timestep (0.05) for openfast, or is there a way to have AMR wind run x-timesteps before it communicates again?

genevievestarke (approved):

PR #90, opened 2024-03-18
Smooth output of FlorisStandin to prevent oscilations in closed-loop response

misi9170:
The `FlorisStandin` is a steady-state model standing in for a the dynamic AMR-Wind model. We (@genevievestarke ) have been finding that this can lead to poor performance in simple closed-loop controllers, which I think is because the plant model essentially has infinite bandwidth, which, when combined with an infinite bandwidth controller (such as the [wind farm power tracking controller implemented in WHOC](https://github.com/NREL/wind-hybrid-open-controller/blob/develop/whoc/controllers/wind_farm_power_tracking_controller.py), which is essentially a proportional controller), appears to lead to a closed-loop system that is marginally stable---see below, which comes from running the [wind farm power tracking example in WHOC](https://github.com/NREL/wind-hybrid-open-controller/tree/develop/examples/wind_farm_power_tracking_florisstandin) with the wind speed at 11m/s, rather than 8m/s.

wf-power-tracking-plot_smoothing0

To solve this issue, this PR "smooths" the output from the `FlorisStandin` model by applying a first order filter to the output power to give the open-loop system a finite bandwidth; this dampens oscillations in the closed-loop system (or removes oscillations entirely). This damping is artificial, in the sense that it is not a direct output from FLORIS; however, any real system with dynamics will have a finite bandwidth, so this update makes the `FlorisStandin` slightly more realistic. The level of smoothing is controllable with the `smoothing_coefficient`, which users can set between 0 (no smoothing) and 1 (as the `smoothing_coefficient` approaches 1, smoothing becomes very heavy).

Using the default `smoothing_coefficient` (0.5) now results in the following (note that Open Loop and Closed Loop control appear similar (both able to achieve the farm setpoint) here because, at 11m/s, there is sufficient resource for the Open Loop controller to achieve the demanded setpoint also. This is not a characteristic of the change to `FlorisStandin`).

wf-power-tracking-plot_smoothing1

## Other notes
This pull request additionally updates the `FlorisStandin` to be compatible with the latest changes in [FLORIS v4](https://github.com/NREL/floris/tree/v4).

### DIFF
--- a/hercules/floris_standin.py
+++ b/hercules/floris_standin.py
@@ -27,7 +27,7 @@ import sys
 from pathlib import Path
 
 import numpy as np
-from floris.tools import FlorisInterface
+from floris import FlorisModel
 from floris.turbine_library import build_cosine_loss_turbine_dict
 
 from hercules.amr_wind_standin import AMRWindStandin, read_amr_wind_input
@@ -119,12 +119,12 @@ def construct_floris_from_amr_input(amr_wind_input):
         turb_dict["power_thrust_model"] = "mixed"
 
         # load a default model
-        fi = FlorisInterface(default_floris_dict)
-        fi.set(
+        fmodel = FlorisModel(default_floris_dict)
+        fmodel.set(
             layout_x=layout_x, layout_y=layout_y, turbine_type=[turb_dict] * len(layout_x)
         )
 
-    return fi
+    return fmodel
 
 
 class FlorisStandin(AMRWindStandin):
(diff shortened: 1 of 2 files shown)


genevievestarke:
What testing script did you use for these figures? I'm running into some floris/flasc version issues with WHOC at the moment, so I might need to update something somewhere

misi9170:
> What testing script did you use for these figures? I'm running into some floris/flasc version issues with WHOC at the moment, so I might need to update something somewhere

I ran the [wind farm power tracking control example in WHOC](https://github.com/NREL/wind-hybrid-open-controller/tree/develop/examples/wind_farm_power_tracking_florisstandin), which generates these figures when it concludes (using [this python file](https://github.com/NREL/wind-hybrid-open-controller/blob/develop/examples/wind_farm_power_tracking_florisstandin/plot_output_data.py))

genevievestarke:
Ah, ok, great, I'm getting the same plots as you now!

genevievestarke (approved):
Looks good, and runs for me with the same results!

PR #92, opened 2024-03-26
updated installation docs

achenry:
I encountered some issues with the installation instructions while installing on Kestrel, fixes are documented here.


genevievestarke:
@misi9170 Was Floris updated again? This seems to be failing all the Floirs tests now.

PR #93, opened 2024-03-26
[BUGFIX] Create outputs folder in examples to avoid error when output file does not exist

misi9170:
A bug was introduced in #84 that means that the AMRWindStandin examples need to be run twice to work.

The issue is that the 
```python
logging.basicConfig(
    level=logging.DEBUG,
    format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s",
    datefmt="%Y-%m-%d %H:%M",
    filename="outputs/log_test_client.log",
    filemode="w",
)
logger = logging.getLogger("amr_wind_standin")
```
at the top of amr_wind_standin.py assumes that the folder outputs/ already exists; however, it is not created until `AMRWindStandin` is instantiated. 

To resolve the issue, we've added the creation of the outputs folder above this code, as well as added it to `emulator.py` in case the emulator is run without the `AMRWindStandin`. Additionally, the `.gitignore` is updated to ignore an log files beginning with `loghercules`, `logstandin`, or `logfloris`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,9 +40,9 @@ hercules/local_amr_wind_demo/sample_copy.nc
 # Some output files to ignore
 t_00*
 logdummy
-loghercules
+loghercules*
 logstandin
-logfloris
+logfloris*
 *echo
 *out-example.json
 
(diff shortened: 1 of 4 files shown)


genevievestarke (approved):
Looks good and runs, I think we can merge it in!

misi9170:
@genevievestarke  Done :)

PR #103, opened 2024-04-12
Change FLORIS to a "regular" requirement

misi9170:
Now that FLORIS v4 has been released, this PR switches the FLORIS installation requirement to the "normal" method for installing packages from PyPI.

I've checked that if I clone this branch; create a fresh environment; and `pip install -e hercules`, the correct version of FLORIS is installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@ REQUIRED = [
     "numpy~=1.20",
     "matplotlib~=3.0",
     "pandas~=2.0",
+    "floris",
     "nrel-pysam~=4.2",
     # "dash>=2.0.0",
     # GUI Stuff
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ REQUIRED = [
     "numpy~=1.20",
     "matplotlib~=3.0",
     "pandas~=2.0",
-    "floris",
+    "floris>=4",
     "nrel-pysam~=4.2",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -38,6 +38,7 @@
     "numpy~=1.20",
     "matplotlib~=3.0",
     "pandas~=2.0",
+    "floris>=4",
```
paulf81, line 41: "~=4.0" would be consistent with other lines, but up to you, FYI ~=4.0 is like >=4, except it won't automatically step up to FLORIS 5, which I think is probably correct

paulf81 (approved):
Looks great @misi9170 , I left only one comment for your consideration

setup.py
misi9170: Yeah, maybe this is my misunderstanding---should the `.0` part be on there? Does `floris~=4` allow version 3? I'm happy to go with your suggestion on this, I'm not super knowledgeable about semantic versioning

paulf81: RIght sorry!  Yes the .0 is needed, floris~=4.0 in my mind means any version of = or greater to the later decimal of the version you specify, so by saying 4.0, 4.1 is ok but not 5.0, if you say 4.1, then 4.0 is also not ok

genevievestarke (approved):
Looks good to me!

PR #104, opened 2024-04-12
Adding solar controller and replacing AMR-Wind standin with FLORIS standin in example 07

brookeslawski:
**1. Solar controller**

Implementing a simple PV plant controller that trims any PV plant power output above the setpoint at each timestep. This is similar to what `pysam` does. The "curtailment" is applied uniformly to the entire PV plant.

Requires an input power setpoint signal, which is read in through the input file field `external_data_file.` 

In the future, this power setpoint might be calculated internally in `hercules` and this controller will need to be modified to handle that input.

**2. FLORIS stand-in**

The AMR-Wind stand-in was causing the errors shown below.
MicrosoftTeams-image (1)

FLORIS was determined to be a better stand-in anyway, and was easy to implement, so the FLORIS stand-in replaced the AMR-Wind stand-in.


example_case_folders/07_amr_wind_standin_and_solar_pysam/test_read_output.ipynb
brookeslawski, line 1: No big changes, just the path leading to the output file and a change in an output variable name

max-power-prediction-OK-pysam.ipynb
misi9170, line 1: Is this file supposed to be included with this PR?

hercules/python_simulators/solar_pysam.py
```diff
@@ -82,12 +100,11 @@ def step(self, inputs):
...
             except Exception:
                 print(k)
-        #### TODO: Check with Brooke about this "except KeyError" line ####
-                # Brooke: I got errors with KeyError so changed it to Exception and it's working
-        # print('model params = ',self.model_params)
 
-        print("sim_time_s = ", inputs["py_sims"]["inputs"]["sim_time_s"])
-        sim_timestep = int(inputs["py_sims"]["inputs"]["sim_time_s"] / self.dt)
+        # sim_time_s = inputs["py_sims"]["inputs"]["sim_time_s"]
+        sim_time_s = inputs["time"]
```
misi9170, line 105: This change (from `inputs["py_sims"]["inputs"]["sim_time_s"]` to `inputs["time"]` is causing the `test_step` test in tests/solar_pysam_test.py to fail. Possibly, the `step_inputs` dictionary in tests/solar_pysam_test.py just need to be updated to add the "time" field?

example_case_folders/07_floris_standin_and_solar_pysam/hercules_controller_input_000.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # capacity: 100 # MW
+
+    initial_conditions:
+
+      power: 25 # MW
+      dni: 1000
+
+controller:
+
+external_data_file: solar_power_reference.csv
```
misi9170, line 59: This file is missing from the repo, I think

misi9170:
@brookeslawski I left a handful of small comments, but perhaps I'm a little confused about the files in the 07_floris_standin_and_solar_pysam. It seems that there is a bit of a mix in there of files that are needed to run this example, and files that maybe are just carry-overs from when it was 07_amr_wind_standin_and_solar_pysam, several of which I think are no longer used.

Did you mean to leave those in? Is the intention that the user could run the example with the AMR-Wind standin, as well as with the FLORIS standin?

max-power-prediction-OK-pysam.ipynb
brookeslawski: Nope, deleting this. Thanks.

hercules/emulator.py
```diff
@@ -145,9 +145,12 @@ def _read_external_data_file(self, filename):
         if "time" not in df_ext.columns:
             raise ValueError("External data file must have a 'time' column")
 
-        # Interpolate the external data according to time
+        # Interpolate the external data according to time.
+        # Goes to 1 time step past stoptime specified in the input file.
         times = np.arange(
-            self.helics_config_dict["starttime"], self.helics_config_dict["stoptime"], self.dt
+            self.helics_config_dict["starttime"],
+            self.helics_config_dict["stoptime"]+(2*self.dt),
```
brookeslawski, line 152: Changed this to create `times` variable that _includes_ the stoptime + one more timestep. `np.arange` is half-open, which does not include `end` and thus was causing errors when hercules was looking for a setpoint at the last timestep, but not finding it even though it was in the input `external_data_file`

hercules/python_simulators/solar_pysam.py
brookeslawski: Fixed! Thanks for pointing this out

example_case_folders/07_floris_standin_and_solar_pysam/hercules_input_000.yaml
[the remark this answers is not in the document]
brookeslawski: In the `batch_script`, I wanted to include the option to run hercules _without_ the pv controller, which uses `hercules_input_000.yaml` as well as the option to run _with_ the pv controller, which uses `hercules_controller_input_000.yaml.` So left both files in there. Is that ok?

brookeslawski:
> Hi @brookeslawski! I left a handful of small comments, but perhaps I'm a little confused about the files in the 07_floris_standin_and_solar_pysam. It seems that there is a bit of a mix in there of files that are needed to run this example, and files that maybe are just carry-overs from when it was 07_amr_wind_standin_and_solar_pysam, several of which I think are no longer used.
> 
> Did you mean to leave those in? Is the intention that the user could run the example with the AMR-Wind standin, as well as with the FLORIS standin?

Hi @misi9170 - Thank you for the review. I've deleted the amr_wind_standin files, which was just the `hercules_runscript_amr_standin.py`

example_case_folders/07_floris_standin_and_solar_pysam/hercules_controller_input_000.yaml
brookeslawski: Added!

example_case_folders/07_floris_standin_and_solar_pysam/hercules_input_000.yaml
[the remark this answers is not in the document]
brookeslawski: README updated to clarify these options

misi9170 (approved):
Thanks for addressing my comments, @brookeslawski ! I've been able to run the example in both controlled and uncontrolled configurations. I made a couple of small changes, hopefully you're happy (but feel free to revert if not):
- Removed one of the `&`s in batch_script.sh, because I found this was causing the script not to terminate properly (and it was my fault that that went in in the first place!)
- Added handling to the test_read_output.ipynb notebook for when running the uncontrolled case

example_case_folders/07_floris_standin_and_solar_pysam/solar_power_reference.csv
genevievestarke, line 1: Just noticing that this file is really long. Do we need all of it, or can we shorten it for example purposes?

example_case_folders/07_floris_standin_and_solar_pysam/test_read_output.ipynb
genevievestarke, line 1: I just edited the results plot so that all of the units were consistent. I don't know if you can see the changes on github though.

genevievestarke (approved):
Looks good to me! I left a few comments that don't really need to be addressed, just some thoughts. Otherwise, good to merge!

example_case_folders/07_floris_standin_and_solar_pysam/solar_power_reference.csv
brookeslawski: I do have a five-timestep version of this file for testing, but I left the longer one in there to be able to demonstrate more of a variation in the setpoint signal. It doesn't take long to run, and you can always change the `stoptime` to be something shorter for quicker tests. The input `stoptime` doesn't need to match the time in this file.

PR #107, opened 2024-05-07
Add external controls for solar PySAM module

misi9170:
Adds the ability to provide a control setpoint to the solar module. The 07_floris_standin_and_solar_pysam example still runs as expected, and I've added a small test to make sure that the control setpoint is adhered to when passed, so I think this is ready for review.

NOTE: I've also saved the AMR wind speed (or stand-in wind speed) on the Hercules main dict for use in controllers that need (an estimate of) the free stream wind speed.


genevievestarke (approved):

brookeslawski:
These changes look good to me!

brookeslawski (approved):
Looks good to me!

Issue #108, opened 2024-05-13
Enable FLORIS to use changing wind direction and thrust control at the same time


genevievestarke:
Currently, FLORIS only allows either yaw control or thrust control. In Hercules, changing wind directions come across as yaw directions to FLORIS, which means that thrust control is not allowed to be implemented at the same time. This causes an error in the code, currently.

paulf81:
I think if you used the "mixed" turbine model you can apply both yaw and thurst:

Here's an example where it's used: https://nrel.github.io/floris/examples/examples_control_types/003_setting_yaw_and_disabling.html

misi9170:
Thanks @paulf81 ! The operation model is already set as `"mixed"`---I think the issue here is that, although we don't really ever intend to by yaw misaligning and derating at the same time, so instances of that are creeping through (possibly because we are trying to apply "realistic" yaw actions, which don't instantaneously match the wind direction, while also trying to derate). I'll look into this more, and override the yaw actions if that's the case.

misi9170:
I think we can close this as long as we accept that in v2, we never want to model yaw misalignments. That would rule out:
- wake steering simulations
- consensus simulations
- simulations where we want to model "realistic" yaw dynamics where turbines are temporarily misaligned as they move to follow changes in wind direction.

I'm OK with that. The clear intention of Hercules v2 is on plant-power-level simulations, i.e. simulations that don't consider wake steering/advanced wind farm controls other than active power control, and where small misalignments are likely to be negligible. 

Of course, handling yaw angles can probably be reintroduced at a later date if we want it in v2.

misi9170:
I will remove `yaw_angles` from Hercules wind simulators.

PR #109, opened 2024-06-02
[BUGFIX] Force `yaw_angles` to zero in `FlorisStandin` when `power_setpoints` passed

misi9170:
@genevievestarke pointed out that an error can occur when using the `FlorisStandin` with a changing wind direction when performing active power control of the turbines. This occurs because the `FlorisStandin` turbines do not perfectly track the wind direction when it changes, which causes a yaw misalignment. FLORIS is not yet capable of simultaneously simulating a yaw misalignment and a demanded power from a turbine, so the simulation errors out. See #108.

The behavior is expected, and is actually tested [here](https://github.com/NREL/hercules/blob/develop/tests/floris_standin_test.py#L159-L160). However, to allow both a changing wind direction and active power control, this pull request ensures that, if a power setpoint is given to a specific turbine, it's yaw angle is set to zero to avoid raising the FLORIS error. 

This is not a perfect solution, because yawing is now possibly unrealistic. If, in future, we have a Floris turbine operation model that can handle [simultaneous yawing and derating](https://github.com/NREL/floris/issues/913), we will switch to that in the `FlorisStandin`.

@genevievestarke , you should now be able to run your example of the hybrid wind/solar/battery plant.

Paul: Added some additional changes in the course of debugging tests:

1. Removed unnecessary explicit install of FLORIS v4 in CI
2. Updated the call to ruff to follow new `ruff check` syntax
3. Added automatic testing of macos in addition to ubuntu-latest so in the future we will know if an issue is os-specific

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -33,7 +33,7 @@ jobs:
       run: |
         # -rA displays the captured output for all tests after they're run
         # See the docs: https://doc.pytest.org/en/latest/reference/reference.html#command-line-flags
-        pytest -rA tests/
+        pytest -vv -rA tests/
     - name: Generate coverage report
       # Run these tests on unit tests only so that we avoid inflating our code
       # coverage through the regression tests
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -27,7 +27,7 @@ jobs:
 #    - uses: pre-commit/action@v3.0.0
     - name: Run ruff
       run: |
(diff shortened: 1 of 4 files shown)


misi9170:
These are failing tests due to a segmentation fault (the tests run fine for me locally). I'll try rerunning them again tomorrow.

misi9170:
After attempting to debug this, we decided it was an upstream bug (possibly from matplotlib). Because it is not affecting our processes locally, we have decided to merge this PR and hope that the bug will be fixed upstream in future. @genevievestarke , I'm going to go ahead and merge.

@paulf81 , please feel free to add any further comments from your attempt to debug the failing tests.

PR #111, opened 2024-08-05
Improving indexing of solar weather file based on dt in input .yaml file

brookeslawski:
This pull request improves the way that `solar_pysam.py` selects data from the input solar weather file. The input solar weather file contains timeseries weather data at a temporal resolution that may or may not be the same as `dt` in the input .yaml file. This pull request ensures that `solar_pysam.py` is selecting weather data at the correct timestamp. For example, if the input solar weather file contains data at every 0.5 second, but `dt = 1.0` s in the .yaml file, the solar module will select data at every second after the timestamp of the first row of data.

If `dt` is not equal to or a multiple (>1) of the temporal resolution in the input solar weather file, a warning will print to the hercules log file.

This pull request does interpolate input data if `dt` is smaller than the temporal resolution in the input solar weather file. This interpolation should happen outside of hercules as part of pre-processing. An example of this pre-processing is included in `example_case_folders/07_floris_standin_and_solar_pysam/read_solar_precursor_data.ipynb`.

I have also improved the handling of timezones and added a warning that timezone must be included in the input solar weather file timestamps.

Also included are the input weather files from the ARM experiment and from the Flatirons campus, which were used for testing.


brookeslawski:
All tests are passing locally. 
image

misi9170:
> All tests are passing locally. image

Thanks @brookeslawski ! See [this pull request](https://github.com/NREL/hercules/pull/109#issuecomment-2258917405) for attempts (which were ultimately unsuccessful) to get these failing tests to clear. For the time being, we'll just ensure the tests pass locally.

brookeslawski:
Per Misha's suggestion, added a `verbose` flag in the SolarPySAM module to toggle on and off the print statements used for debugging. Re-ran tests and they all pass locally.

hercules/emulator.py
```diff
@@ -238,7 +238,7 @@ def receive_amrwind_data(self):
         # Assign Py_sim outputs
         if self.main_dict["py_sims"]:
             self.main_dict["py_sims"]["inputs"]["available_power"] = sum(turbine_power_array)
-            print("sim_time_s_amr_wind = ", sim_time_s_amr_wind)
+            # print("sim_time_s_amr_wind = ", sim_time_s_amr_wind)
```
brookeslawski, line 241: cleaning up print statements that I had added long ago

genevievestarke (approved):
This all looks good! I think down the line we'll want to put 'verbose' as a global variable, like dt, but I think this works for now!

PR #123, opened 2024-11-20
Standardize Examples

paulf81:
@genevievestarke @misi9170 @abhineet-gupta @ZackTully @brookeslawski 

In this pull request, I'd like to go through and standardize the examples.  I'd very much welcome input into the project as I got along!  Putting this branch on origin so folks will feel free to commit directly to it, or if rather submit pull requests.

## High level goals:

- [x] All readme files are `README.md`, explain the example and say how to run the example locally and on kestrel
- [x] Runscripts standardized to: (1) run_script.sh - for running locally, (2) sbatch_script.sh (which calls run_script.sh) and runs on kestrel (exception to pattern if example only runs on kestrel)
- [x] Clean up post analysis scripts

## Setting aside KESTREL examples

We've decided to not include the criteria that the AMR-WIND/KESTREL examples are fully included cleaned up.  The work of this is now flagged in ISSUE #131 

## Activating CONDA

Found an approach that should enable activating conda on local machines in bash scripts as well as on kestrel.  Adding these lines to run_scripts:

```
# Determine the base path for Conda initialization
if [ -f "/home/$USER/anaconda3/etc/profile.d/conda.sh" ]; then
    # Common path for Anaconda on Linux
    CONDA_PATH="/home/$USER/anaconda3/etc/profile.d/conda.sh"
elif [ -f "/Users/$USER/anaconda3/etc/profile.d/conda.sh" ]; then
    # Common path for Anaconda on macOS
    CONDA_PATH="/Users/$USER/anaconda3/etc/profile.d/conda.sh"
elif [ -f "/opt/anaconda3/etc/profile.d/conda.sh" ]; then
    # Alternative system-wide installation path
    CONDA_PATH="/opt/anaconda3/etc/profile.d/conda.sh"
elif [ -f "opt/miniconda3/etc/profile.d/conda.sh" ]; then
    # Alternative system-wide installation path
    CONDA_PATH="/opt/miniconda3/etc/profile.d/conda.sh"
elif command -v conda &> /dev/null; then
    # If conda is in PATH, use the which command to find conda location
    CONDA_PATH=$(dirname "$(which conda)")/../etc/profile.d/conda.sh
else
    echo "Conda installation not found. Please ensure Conda is installed and in your PATH."
    exit 1
fi
```

## Specific goals

- [x] Example 5 cleaned up
- [x] Example 5 runs locally
- [x] Example 6 cleaned up
- [x] Example 6 runs locally
- [x] Example 7 cleaned up
- [x] Example 7 runs locally
- [x] Example 7 solar produces power
- [x] Example 8 cleaned up
- [x] Example 8 runs locally
- [x] Example 8 vis output
- [x] Example 10 cleaned up
- [x] Example 10 runs locally
- [x] Renumber 10 back to 9

## Clean up after above

- [x] Re-number examples
- [x] Make sure all docs build


genevievestarke:
Just as a comment, using the python3 command doesn't work for me, so I always change it to python in the run scripts. Thoughts?

ZackTully:
I found an issue in the examples that use either `floris_standin` or `amr_wind_standin`. Both standin simulations require a csv input file generated by `hercules/tools/generate_amr_standin_data.py` but this input file isn't created in `run_script.sh` or the hercules runscript `.py` files. If you already have the `XXX_standin_data.csv` file in the example folder then it runs just fine but somebody who just installed hercules won't have these data files the first time they run an example.

A potential solution is: in `06_amr_wind_standin_and_battery/batch_script.sh`, there is a call to `generate_amr_standin_data.py`, which generates the required standin data csv. However, `generate_amr_standin_data.py` is hardcoded to save the file in example 06 so this won't work for other examples in its current state.

paulf81:
> Just as a comment, using the python3 command doesn't work for me, so I always change it to python in the run scripts. Thoughts?

This is happened to me too.  In my case, `which python` returns the correct answer of the one in the conda env, while `which python3` is the system one.  I don't really get the python3 command so happy to just use python

misi9170:
After merging in develop (with #125), tests are still failing but for reasons that we can now work on addressing.

paulf81:
> @paulf81 I think I've fixed 10---if you could pull again and verify, that would be great.
> 
> The main catch was some code that was introduced in #118 (specifically, [this line](https://github.com/NREL/hercules/blob/e910d4ee6baffef3d4cb0dce29e9ea29d7e46e49/hercules/py_sims.py#L78)) that assumed the `SolarPySAM` module for solar simulations; however, that line doesn't work with the `SimpleSolar` module. I've added a `try` / `except` clause to deal with it for now, but the root issue is that `SimpleSolar` and `SolarPySAM` don't return similar dictionaries in their `return_outputs()` methods, which they probably should?

Example 10 works for me now, thank you @misi9170 !

paulf81:
I added a plot output scripts to example 10:

image

paulf81:
> Just as a comment, using the python3 command doesn't work for me, so I always change it to python in the run scripts. Thoughts?

Removed all `python3` -> `python`

brookeslawski:
Example 7 is now fixed. This example appeared as if it were producing no power because the input solar data file had such a low DNI, that the power was close to zero. I trimmed the input data file `Flatirons_solar_data_highDNI.csv` to shift the simulated period to later in the day when the DNI exceeds 900 W/m2 so the solar power output is much greater than zero. In the plot on the left, solar power output appears to be constant only because 20 seconds is not long enough to see a meaningful increase in DNI (e.g. irradiance only changes by 0.2 W/m2).

image

example_case_folders/07_floris_standin_and_solar_pysam/hercules_input_Flatirons.yaml
```diff
@@ -42,8 +42,8 @@ py_sims:
 
     py_sim_type: SolarPySAM
     verbose: True # flag to print debugging statements
-    weather_file_name: Flatirons_solar_data_precursor-tztest.csv
-    system_info_file_name: 100MW_1axis_pvsamv1.json
+    weather_file_name: inputs/Flatirons_solar_data_highDNI.csv
```
brookeslawski, line 45: trimmed input solar data file to shift the simulation to a time period with higher DNI so that solar power output is much greater than zero

brookeslawski (commented):
I reviewed Example 7 only, but it looks good to me and runs locally without issues

misi9170:
Just merged develop back in to resolve any possible conflicts coming from changes in #129

.gitignore
```diff
@@ -35,8 +35,9 @@ hercules/local_amr_wind_demo/sample_copy.nc
 #Ignore csv files
 *.csv
 !tests/test_inputs/*.csv
-!example_case_folders/10_heterogeneous_wind/floris_standin_data.csv
-!example_case_folders/10_heterogeneous_wind/wind_power_reference_data.csv
+!example_case_folders/10_heterogeneous_wind/inputs/floris_standin_data.csv
+!example_case_folders/10_heterogeneous_wind/inputs/wind_power_reference_data.csv
+!example_case_folders/07_floris_standin_and_solar_pysam/inputs/Flatirons_solar_data_precursor-tztest.csv
```
misi9170, line 40: Seems like we also need to add `'!example_case_folders/07_floris_standin_and_solar_pysam/inputs/Flatirons_solar_data_highDNI.csv` here, if I'm understanding the issue correctly?

misi9170:
I've now been able to run all examples listed in the PR description successfully. However, I do want to make a note that this PR makes changes to other examples as well (1--4) without actually running them. Just noting in case we come back to this point in future while tracing back changes to examples 1--4.

misi9170:
@paulf81 , I found an extra batch_script.sh in the 06 example, and removed it... then found that it was used the the check examples workflow, so changed that to run run_script.sh instead... and then found that that didn't work well in the github actions (it just seemed to hang).

I've since reverted those changes so that the check examples will run, but do you know why we can't use the run_script.sh from example 06 (I have a feeling you've probably come across this yourself, since you left the 06 batch_script.sh in place)? It feels annoying to have to leave the batch_script.sh in example 06, and arguably renders the check examples action a bit impotent if it doesn't check the example using the same approach we would actually use to run it?

paulf81:
hi @misi9170 , it's ringing a distant bell, but I think it's a hack.  I have some new ideas on how to do the check examples action, maybe we clean this one out, and I'll start a new PR to check all the examples 5-9 using our runscript.

paulf81:
ok @misi9170 check examples is now removed, I'll open a new PR/issue to set up check examples in a fashion that runs the examples how we would run them.

I think this ones ready if you don't mind a final review

misi9170 (approved):
Thanks @paulf81 !

genevievestarke (approved):
All the examples run locally for me, I'm good to merge in when you're ready!

PR #125, opened 2024-11-21
[BUGFIX] Specify previous OS for ubuntu tests.

misi9170:
After various debugging attempts, it seems that tests are failing only on `ubuntu-latest` (which is currently 22.04, see [here](https://github.com/actions/runner-images)). This pins the testing matrix back to the previous ubuntu OS, 20.04.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: False
       matrix:
         python-version: ["3.9", "3.10", "3.11"]
-        os: [ubuntu-latest] #, macos-latest, windows-latest]
+        os: [ubuntu-20.04] #, macos-latest, windows-latest]
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/continuous-integration-workflow.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: False
       matrix:
         python-version: ["3.9", "3.10", "3.11"]
-        os: [ubuntu-latest] #, macos-latest, windows-latest]
+        os: [ubuntu-20.04] #, macos-latest, windows-latest]
```
misi9170, line 13: Strictly, it's only this line that actually had to change to have all tests pass (for some reason, the check-working-examples workflow was still running correctly). However, for consistency, I've switched all instances of `ubuntu-latest` to `ubuntu-20.04`.

paulf81: Agree with this

paulf81 (approved):
Thank you @misi9170 !

genevievestarke (approved):
Looks good, I'll merge in!

PR #129, opened 2024-12-02
Add regression testing

misi9170:
Hercules doesn't currently have any regression tests for the simulation modules, which makes it hard to tell when unintentional changes happen in new code developments. This PR adds a regression_tests/ subdirectory to tests/ and begins to build out regression tests for the various simulation modules.

Regression tests developed (crossed-out entries will not be included in this PR, for the reason listed):
- [x] ~`AMRStandin`~ (not likely to use as a simulator---just as a base class)
- [x] `FlorisStandin`
- [x] `SolarPySAM`
- [x] ~`SimpleSolar`~ (no dynamics implemented here---just returns a steady output)
- [x] `SimpleBattery`
- [x] `LIB`
- [x] `ElectrolyzerPlant`


misi9170:
tests are currently failing, likely due to a mismatch in the version of electrolyzer. I'll look into this.

.github/workflows/continuous-integration-workflow.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -e ".[develop]"
-        pip install git+https://github.com/NREL/electrolyzer.git
```
misi9170, line 25: Allows us to use the updated inputs to the electrolyzer in the regression tests.

genevievestarke (approved):
Looks good to me, and runs locally!
We might also want to look at regression test for running the pysams within the emulator framework, but that can be a later issue

PR #132, opened 2024-12-05
Update installation instructions

paulf81:
This small PR updates the installation codes and instructions to make a few improvements to current practice.  Main current changes are:

## Changes
- Changes SEAS reference from main to v1
- Removes unnecessary pysam installation instruction since this is now a dependency
- Cleans up installation instructions


docs/install_local.md
```diff
@@ -85,22 +93,20 @@ git fetch --all
...
+```
+bash run_script.sh
+```
+
+## Running manually
 
-To run locally using a standin for AMR-Wind, launch 3 separate 
+To run locally without the included runscript using a standin for AMR-Wind, launch 3 separate 
 terminals. In each, `conda activate` your hercules environment (`conda 
 activate hercules`).
```
misi9170, line 112: I can't actually comment on the lines below this, because this PR doesn't change them, but they seem to assume the user has already declared a `HELICS_PORT` variable?

paulf81: Yeah, I'm not positive about this whole section.  In the same way as the runscript in each folder might be a little different, the manual version might also be a little different and the best approach would be just to enter the lines appearing in the runscript as they appear.  Possibly in 3 terminals, or using the & to unblock, maybe just shorten this section to something more like "to be sure, it's possible to enter the commands in the runscript directly..."?

misi9170: I think it makes sense to encourage people to use the run_script.sh . So, perhaps those instructions should come first. But then, maybe it is worth taking the example of one of the run_script.sh , and talking through what each line is doing (and explain that it is possible to run this from three separate terminals, if that's what the user would prefer to do)?

paulf81: Coming back to this discussion, it seems to me the much improved run scripts make this section a bit moot.  Being bash scripts the user is free to enter the statements line by line so I think I'll just say something to the effect of each run script illustrates how to run manually

paulf81:
@genevievestarke  and @misi9170  think this one's ready to go

genevievestarke (approved):

PR #139, opened 2025-02-19
Update PySAM to v6.0.0

brookeslawski:
Hercules had previously used PySAM v4.2.0, so this PR upgrades the PySAM version to the latest, which is v6.0.0.

This upgrade involved the following changes:
- changing the name of two variables that the PySAM model requires as inputs
- replacing the solar data input file with a modified version of `inputs/Flatirons_solar_data_highDNI.csv` called `inputs/Flatirons_solar_data_sunset.csv`, which shows more dynamic fluctuations in the solar farm power output - this is better for testing the solar module
- updating documentation for installation to refer to v6.0.0
- updating results in tests to match PySAM v6 values - there was a small change in the temperature model from v5 to v6 that could explain the difference in output power (e.g. 32.19801938915135 vs 32.17650018440107 MW)


scripts/activate_conda.sh
```diff
@@ -22,6 +22,7 @@ else
     echo "Conda installation not found. Please ensure Conda is installed and in your PATH."
     exit 1
 fi
+echo "$CONDA_PATH"
```
brookeslawski, line 25: Not sure if I'm the only one, but HOPP has trouble finding my conda env so I thought this would be nice to add in case others have that issue too. Just helps with debugging

genevievestarke (approved):

PR #140, opened 2025-02-20
Add dof1_model for long term simulation

abhineet-gupta:
Add dof1_model for long duration wind farm simulations.


paulf81:
@abhineet-gupta should `notes.md` be deleted before merging?  It looks like notes for development but not needed in final right?

abhineet-gupta:
> @abhineet-gupta should `notes.md` be deleted before merging? It looks like notes for development but not needed in final right?

Yes, that should be deleted.

paulf81 (approved):

PR #135, opened 2024-12-12
Long Run Hercules

paulf81:
# Long Run Hercules

Opening this draft pull request, with the branch on origin, to facillitate the development of long-running Hercules.   @abhineet-gupta , you and I can either commit direct to this branch or else submit PRs to the branch from our forks.  

We can further use this PR to track what we want to do and what we consider to be done.  We will likely edit this front matter a lot as we go.

# Items to do

- [ ] Set up new long-term running skeleton as alternative to amr-wind/floris-standin
- [ ] Implement 2-DOF turbine model
- [ ] Add FLORIS wakes
- [ ] Connect to inflow reads
- [ ] Make a wind-only example
- [ ] Make a wind+solar example
- [ ] Documentation

# Tagging people
@abhineet-gupta @genevievestarke @misi9170 @brookeslawski @christiannvaughn

PR #144, opened 2025-04-03
Add solar to long run hercules for longterm simulation of wind and solar

brookeslawski:
Long run Hercules is being developed in PR #135 to enable long duration runs (on the order of 1 year). The `feature/long_run` branch only has wind modeling included so this PR adds solar modeling in a new example (Ex. 13).

To enable solar modeling in long run Hercules, a substantial speed-up is required compared to the computational performance of the existing solar python simulator, which uses PySAM's Detailed PV model. The changes in this PR include:

- Addition of a user-specified input option to run either PySAM's Detailed PV model (`PySAM.Pvsamv1`) or the faster `pvwatts` model to enable long-duration runs.
- Edits to plotting script to also plot solar power output on a 2nd figure.
- Bug fix: Ex 07 and `solar_pysam.py` with `pvsam` wasn't backtracking like we had thought, so I corrected this in the input json file and in the solar pysam tests.

On my local machine, I'm seeing a 150x real time speedup in this example (with wind and solar) compared to a 400x real time speedup of the wind-only example (Ex. 10).

- [x] Paul to check the speed-up comparison between Examples 10 and 13 when he runs it
- [x] Add pvwatts tests
- [x] Undo user-specific changes to `activate_conda.sh`
- [x] Add explanation of new input flag in `solar_pv.md`. Anything else to do in documentation?
- [x] Paul to create a separate issue to allow usage by conda/mamba/uv users


scripts/activate_conda.sh
brookeslawski, line 1: This is something I change to run hercules because I use mamba instead of conda. I'll change this back before merging.

hercules/python_simulators/solar_pysam.py
```diff
@@ -149,13 +161,14 @@ def step(self, inputs):
...
-                [data["SRRL BMS Wind Speed at 19' (m/s)"]],
-                [data["SRRL BMS Dry Bulb Temperature (°C)"]],
+                [self.data[row_index,self.col_dict['dni_col']]],
+                [self.data[row_index,self.col_dict['dhi_col']]],
+                [self.data[row_index,self.col_dict['ghi_col']]],
+                [self.data[row_index,self.col_dict['ws_col']]],
+                [self.data[row_index,self.col_dict['temp_col']]],
             ]
         )
+        # print('weather_data = ', weather_data)
```
paulf81, line 171: Delete in final

scripts/activate_conda.sh
paulf81: Good to know, we could probably try to make this script more one-size-fits-all in the future

paulf81:
@brookeslawski , I cleaned `example_case_folders/13_WindSolarSimLongTerm/resample_solar_history.ipynb` a little bit.  I was wondering, do we need to keep the second half?  It seems like debugging stuff that can now come out?

paulf81:
`example_case_folders/13_WindSolarSimLongTerm/inputs/Flatirons_solar_data_sunset_1s.csv` is 4.6M.  @misi9170 , I was thinking this should not be included but since it's already included I'm not sure deleting saves us anything unless we do some backward cleaning.  It's generated by `example_case_folders/13_WindSolarSimLongTerm/resample_solar_history.ipynb` so I don't think it has to be included but not sure what do here

paulf81:
> `example_case_folders/13_WindSolarSimLongTerm/inputs/Flatirons_solar_data_sunset_1s.csv` is 4.6M. @misi9170 , I was thinking this should not be included but since it's already included I'm not sure deleting saves us anything unless we do some backward cleaning. It's generated by `example_case_folders/13_WindSolarSimLongTerm/resample_solar_history.ipynb` so I don't think it has to be included but not sure what do here

similar comment on `wind_input.csv`

paulf81:
@brookeslawski  is the plot I make at the end reasonable?  it seems like the irradiance goes to 0 but still there is power generated:

image

paulf81:
For speed I'm getting 391x with solar, 1571x without

misi9170:
@paulf81 , regarding those large data files---actually, I think deleting them at this stage would keep them out of the main history. Because we squash and merge these PRs into develop, there is just a single commit to develop that only keep the final changes. So, I'm pretty sure that anyone that pulls develop (and eventually, main) _won't_ end up with those large files in their history, as long as the never make it onto develop/main. 

(I think) they only live on on github, where someone could actually look at this PR, find those specific commits that added/removed the files (commits which are not actually on the develop commit history), and get the files from those commits directly.

scripts/activate_conda.sh
misi9170, line 1: Looks like these changes are user-specific and perhaps aren't meant to go in?

hercules/python_simulators/solar_pysam.py
```diff
@@ -77,10 +84,26 @@ def __init__(self, input_dict, dt):
         self.dni = input_dict["initial_conditions"]["dni"]
         self.aoi = 0
 
+        # create pysam model
+        system_model = pvwatts.new()
+        system_model.assign(self.model_params)
```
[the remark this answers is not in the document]
paulf81: isn't this what happens on line 64?

brookeslawski:
> @brookeslawski is the plot I make at the end reasonable? it seems like the irradiance goes to 0 but still there is power generated:
> 
> image

I'm only plotting on component of irradiance (DNI) so most likely, at around sunset, DNI goes to zero, but the diffuse component is still present, which contributes to non-zero power output. I'll update the y-axis label in the plot to clarify that it's just DNI.

hercules/python_simulators/solar_pysam.py
brookeslawski: Done

scripts/activate_conda.sh
brookeslawski: Yeah, see my response to Paul's comment above. They have now been removed, but we are creating an issue to prevent the need for these changes when using mamba instead of conda.

tests/regression_tests/solar_pysam_regression_test.py
brookeslawski, line 1: These changes are the result of changing the tracking mode to backtracking

tests/test_inputs/100MW_1axis_pvsamv1.json
```diff
@@ -236,7 +236,7 @@
 	"subarray1_mod_orient": 0,
 	"subarray1_nmodx": 7,
 	"subarray1_nmody": 2,
-	"subarray1_backtrack": 0,
+	"subarray1_backtrack": 1,
```
brookeslawski, line 239: setting this to 1 means the system will backtrack, as desired

example_case_folders/07_floris_standin_and_solar_pysam/hercules_input_Flatirons.yaml
```diff
@@ -41,15 +41,15 @@ py_sims:
...
     py_sim_type: SolarPySAM
+    pysam_model: pvsam #pvsam or pvwatts
     verbose: True # flag to print debugging statements
     weather_file_name: inputs/Flatirons_solar_data_sunset.csv
     system_info_file_name: inputs/100MW_1axis_pvsamv1.json
     lat: 39.7442
     lon: -105.1778
     elev: 1829
-    target_system_capacity: 130000
+    target_system_capacity_kW: 130000
```
brookeslawski, line 51: This got lost somewhere in the PRs, but we should include units in this variable name.

misi9170:
Tests were not running because `ubuntu-20.04` was specified for testing, and this is apparently being retired:
image

I've now reverted the change introduced in #125 so that tests run again on `ubuntu-latest`, which appears to fix the issue. At the time of this comment, `ubuntu-latest` is 24.04.

paulf81 (approved):
Test passes and new example plot is great, thanks @brookeslawski !

genevievestarke (approved):

Issue #96, opened 2024-04-02
Add plant output variable


genevievestarke:
Add a variable (or group of variables) that gives information on the output of the plant in electricity, hydrogen, or both.

genevievestarke:
Put this in the electrical balance class

genevievestarke:
Partially addressed in PR #118 for including whole plant in available_power variable

PR #149, opened 2025-05-13
Provide options for grid charging and charging commands for batteries, electrolyzers

misi9170:
Adds the following capabilities:
- Electrolyzer can charge following a commanded power consumption
- Electrolyzer and battery may charge from grid, if set up to do so

To do:
- [x] Implement main changes to source code
- [x] Check all existing tests pass/update tests so to test same functionality
- [ ] Add tests for new functionality
- [ ] Check all running examples still run and produce the same results

### DIFF
--- a/docs/battery.md
+++ b/docs/battery.md
@@ -11,6 +11,7 @@ Battery parameters are defined in the hercules input yaml file used to initializ
 - `charge_rate`: [kW]
 - `max_SOC`: between 0 and 1
 - `min_SOC`: between 0 and 1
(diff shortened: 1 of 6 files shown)


misi9170:
@genevievestarke , we could also use this opportunity to add a better "grid output power" output to the main dictionary and help address #96 . However, we'll need to add an output to the electrolyzer for power consumed if we want to do that.

genevievestarke:
@misi9170 I added a plant output variable for electricity and hydrogen!

PR #150, opened 2025-05-16
Adding poa irradiance as output from solar python simulator

brookeslawski:
This PR extracts the plane of array (poa) irradiance (`poa`) as an output from pysam in the solar python simulator. This variable enables easy plotting of the irradiance input to the solar farm, which is the primary determinant of solar farm power output.

POA irradiance is the sum of the direct (DNI), diffuse (DHI), and reflected (GHI) irradiance components as they are projected onto the tilted plane of the solar panel. Up to this point, plotting scripts of the hercules outputs sometimes included `dni` (direct normal irradiance), but this is just one component of irradiance. POA irradiance a single value that encapsulates the incoming solar irradiance to the solar farm.

I created this change in this branch so that it would apply to both `pvsamv1` and `pvwatts` models in the solar python simulator.


paulf81 (approved):
This looks great, thanks @brookeslawski !

paulf81:
@genevievestarke  and @misi9170 , are you ok if we merge this one into long_run?

paulf81:
Going to merge into long_run, figuring that one still has a review to go can tackle all as one

PR #153, opened 2025-06-18
[BUGFIX] Check whether PySims exist before calculating outputs

misi9170:
#149 introduced methods for computing outputs from PySims, but these methods assumed that all Hercules simulations contains PySims. As it stands, running Hercules with only a wind plant means that `Emulator.main_dict["py_sims"]` is `None`, and trying to access subdictionaries caused wind-only simulations in WHOC to break.

This PR addresses the bug by first checking if PySims exist before trying to calculate outputs from them.

Note that this may not be needed in #135 , when wind becomes another type of PySim.

### DIFF
--- a/hercules/emulator.py
+++ b/hercules/emulator.py
@@ -42,9 +42,8 @@ class Emulator(FederateAgent):
         # Initialize time # TODO - does this belong in 'inital conditions' instead?
         if self.main_dict["py_sims"]:
             self.main_dict["py_sims"]["inputs"]["sim_time_s"] = 0.0
-
-        self.main_dict["py_sims"]["inputs"]["plant_outputs"] = {}
-        self.main_dict["py_sims"]["inputs"]["plant_outputs"]["electricity"] = 0.0
+            self.main_dict["py_sims"]["inputs"]["plant_outputs"] = {}
+            self.main_dict["py_sims"]["inputs"]["plant_outputs"]["electricity"] = 0.0
 
         # Initialize plant and grid variables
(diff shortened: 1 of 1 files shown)


misi9170:
The issue and fix can be tested by running the 08_floris_standin_only/ example. 
- On 38da10e (prior to the merger of #149), the example runs to completion
- On 74dfee2 (the merger of #149, currently the head of develop), the example fails
- On this big fix branch, the example runs to completion and generates outputs that match the prior behavior.

The failing example generates the following error in outputs/loghercules.log:
```
Running Hercules with helics_port 32000
Traceback (most recent call last):
  File "/Users/msinner/projects/herc_whoc/hercules/example_case_folders/08_floris_standin_only/hercules_runscript.py", line 24, in <module>
    emulator = Emulator(controller, py_sims, input_dict)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/msinner/projects/herc_whoc/hercules/hercules/emulator.py", line 46, in __init__
    self.main_dict["py_sims"]["inputs"]["plant_outputs"] = {}
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```
This occurs because there are no PySims, so `Emulator.main_dict["py_sims"]` is `None` (and so trying to access the `"inputs"` key causes a failure). The fix is to simply check if `Emulator.main_dict["py_sims"]` is not None prior to accessing the `"py_sims"` subdictionary.

genevievestarke (approved):

PR #165, opened 2025-10-23
Add grid status tool

paulf81:
This PR adds a script for automatically downloading and readying LMP data from grid status into Hercules.


examples/grid/grid_status_download_example.py
```diff
@@ -0,0 +1,59 @@
+# Example version of the gridstatus_download.py script
+# Download 2 days of data for the OKGE.FRONTIER location
+# Download both real time and day ahead data
+
```
ejsimley: Consider a note about needing an API key to run this example

docs/gridstatus_download.md
```diff
@@ -0,0 +1,54 @@
...
+- Power system reliability data
+
+## API Key Setup
+
+To use this script, you'll need an API key from Grid Status:
+
+1. Visit [Grid Status API Settings](https://www.gridstatus.io/settings/api)
+2. Sign up for an account if you don't have one
+3. Generate an API key
+4. Set the API key as an environment variable or configure it in your Grid Status client
```
ejsimley: Mentioning the specific steps for setting as an environment variable would be helpful. The README file here clarified it for me: https://github.com/gridstatus/gridstatusio

docs/gridstatus_download.md
```diff
@@ -0,0 +1,54 @@
...
+   uvx --with gridstatusio --with pyarrow python grid_status_download.py
+   ```
+
+## Parameters
+
+- **dataset**: The Grid Status dataset identifier (e.g., "spp_lmp_real_time_5_min")
+- **start**: Start date in YYYY-MM-DD format
+- **end**: End date in YYYY-MM-DD format  
+- **filter_column**: Column to filter on (usually "location" to select a node)
+- **filter_value**: Value to filter by (should be the name of a node of interest, eg "OKGE.FRONTIER")
```
ejsimley: Similarly, is there a list of nodes somewhere that could be referenced here?

examples/grid/grid_status_download_example.py
```diff
@@ -0,0 +1,59 @@
+# Example version of the gridstatus_download.py script
+# Download 2 days of data for the OKGE.FRONTIER location
+# Download both real time and day ahead data
```
ejsimley: Somewhere, a short overview of the two example scripts (this one and "process_results.py") would be helpful. For example, does this script download the raw files and then "process_results.py" puts them in the format required by hercules?

ejsimley (changes requested):
This worked for me! I left some suggestions where I thought some more information would be helpful in the documentation. A couple other comments to consider:

- Should the `uv` package be added to the requirements?
- I thought it might be cleaner to just have the grid status download script in the `examples/grid` folder instead of also having `grid_status_download.py` in `hercules/grid`. They seem very similar, so might be less confusing to just have the example script that users can modify and move to the directory they're working in and then leave the modules with the utilities functions in the `hercules/grid` folder.

docs/gridstatus_download.md
```diff
@@ -0,0 +1,54 @@
...
+- **QUERY_LIMIT**: Maximum number of rows to download (default: 20,000.)  Included to avoid accidentally using too much of account limit.
```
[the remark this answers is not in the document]
paulf81: I think this depends on your tier, although every account has some limit.  The documentation just mentions if you just happen to forget to specify a node and download every node in a dataset you might inadvertently burn a lot of your allotment so having a sensible limit is a good failsafe.  Could add more language to this note to make this more clear?

docs/gridstatus_download.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Output
+
+The script saves the downloaded data as a feather file (`.ftr`) with:
+- Cleaned column names (removes columns unused by Hercules)
+- Renamed time column to `time_utc`
+- Filename format: `gs_{dataset}_{start}_{filter_value}.ftr`
+
+The feather file can be used now to provide LMP information to Hercules runs.
```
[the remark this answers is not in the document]
paulf81: Ok think I've reorganized this a bit to discuss this distinction, thanks for raising!

docs/gridstatus_download.md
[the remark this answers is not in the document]
ejsimley: Either way is fine. True that depending on if someone has a paid account or free account they'd have different limits.

misi9170:
Regarding the duplication of code between the examples and source code: I do agree that we don't need both grid_status_download.py and grid_status_download_example.py as they are.

An option would be to create a function `grid_status_download()` that performs essentially the code that is in grid_status_download.py (and returns `df`); then, the example could simply call this function and write the dataframe to a feather. I guess the question is whether that function would actually be useful, or whether the user will almost always need to modify the code. If they're going to anyway need to modify it, I would just scrap grid_status_download.py and leave the example script.

paulf81:
> I don't think I've used `uvx` before, so when I ran this command `uvx --with gridstatusio --with pyarrow python grid_status_download_example.py`, I had to go back and pip install uv. So thought it would be convenient to install it as part of the Hercules installation. At least for someone (like me) who pip installed hercules in a fresh conda environment, I didn't have `uv` installed yet. But maybe since `uv` is only required to run the grid status script and not the main hercules package, it doesn't need to be listed as a dependency?

hi @ejsimley , I added some docs and links on installing uv to the docs, does that look good?

ejsimley:
> > I don't think I've used `uvx` before, so when I ran this command `uvx --with gridstatusio --with pyarrow python grid_status_download_example.py`, I had to go back and pip install uv. So thought it would be convenient to install it as part of the Hercules installation. At least for someone (like me) who pip installed hercules in a fresh conda environment, I didn't have `uv` installed yet. But maybe since `uv` is only required to run the grid status script and not the main hercules package, it doesn't need to be listed as a dependency?
> 
> hi @ejsimley , I added some docs and links on installing uv to the docs, does that look good?

Yeah, that looks good to me.

misi9170:
To do:
- [ ] Decide on creating download function (move to grid_utilities.py?) vs only having downloading code in example @paulf81 .
- [ ] Add tests for `generate_locational_marginal_price_dataframe()` @misi9170

misi9170:
@paulf81 I think we're now good here.

PR #171, opened 2025-11-04
Doc/docstring

paulf81:
Small (AI-driven) pull request just to do a sweep and add in any missing docstrings and enforce the google-style on all existing docstrings.


misi9170:
Hi @paulf81 , I don't see any issues here. However, should we be adding something to the ruff rules (or similar) to enforce certain styling going forward? I can see this format not being strictly followed by people (including myself) going forward, and this seems like an opportunity to set some standards around this.

paulf81:
> Hi @paulf81 , I don't see any issues here. However, should we be adding something to the ruff rules (or similar) to enforce certain styling going forward? I can see this format not being strictly followed by people (including myself) going forward, and this seems like an opportunity to set some standards around this.

We could include pydocstyle in ruff linting (option D) I think is one option

misi9170 (approved):

PR #173, opened 2025-11-05
Reconfigure wind models

paulf81:
Replaces #172 

This pull request reorganizes the wind farm simulation component in a few ways:

Combines what had been two separate classes (Wind_MesoToPower and Wind_MesoToPowerPrecomFloris) into one (WindFarm) class to capitalize on a lot of repeated code outside the wake modeling

Adds two new component types:
a) Wind_MesoToPowerNoAddedWakes uses the wind speeds in the input without adding wakes (added as 3rd option to WindFarm)
b) WindFarmSCADAPower implemented as new class because more substantially different, passes through power in SCADA data but does allow limiting by setpoints (although worth noting power in SCADA may already be de-rated)

Besides that adds relavent tests, examples and docs

### DIFF
--- a/hercules/plant_components/wind_farm_scada_power.py
+++ b/hercules/plant_components/wind_farm_scada_power.py
@@ -274,8 +274,7 @@ class WindFarmSCADAPower(ComponentBase):
 
 
 class TurbineUpdateModelVectorizedSCADA:
-    """Vectorized wind turbine update model for power output simulation.
-    """
+    """Vectorized wind turbine update model for power output simulation."""
 
     def __init__(self, dt, initial_scada_powers):
         """Initialize the vectorized turbine model.
@@ -293,6 +292,8 @@ class TurbineUpdateModelVectorizedSCADA:
 
         # Initialize the previous powers for all turbines
         self.prev_powers = initial_scada_powers.copy()
+
+        print("Filtering not yet implemented for SCADA-based turbine model. Use with caution.")
 
     def step(self, scada_powers, power_setpoints):
         """Simulate a single time step for all wind turbines simultaneously.


misi9170:
Thanks @paulf81 ! I'll get this reviewed shortly, possibly Monday. I'm not sure I totally follow on the formatting---from what I see, #191 does not actually contain any formatting changes and the [6 files that are flagged in the checks here](https://github.com/NREL/hercules/actions/runs/20380839045/job/58570670101?pr=173) are all ones that have been modified in this PR. Possibly just something weird to do with when the branches were made? In any case, I think that they should be cleared up by just modifying a few files already changed in this PR. I can check that quickly now.

paulf81:
> Thanks @paulf81 ! I'll get this reviewed shortly, possibly Monday. I'm not sure I totally follow on the formatting---from what I see, #191 does not actually contain any formatting changes and the [6 files that are flagged in the checks here](https://github.com/NREL/hercules/actions/runs/20380839045/job/58570670101?pr=173) are all ones that have been modified in this PR. Possibly just something weird to do with when the branches were made? In any case, I think that they should be cleared up by just modifying a few files already changed in this PR. I can check that quickly now.

Ah, quite possibly I got myself confused.  I was thinking I was seeing format changing files that would not have been otherwise changed so forced those changes into #191. In any case if we finish that one and merge and pull into here we'll be on right path.

examples/02c_wind_farm_realistic_inflow_direct/hercules_input.yaml
```diff
@@ -0,0 +1,42 @@
...
+  log_channels:
+    - power
+    - wind_speed_mean_background
+    - wind_speed_mean_withwakes
+    - wind_direction_mean
+    - turbine_powers
+    - wind_speeds_withwakes
+    - wind_speeds_background
+    - turbine_power_setpoints
+  floris_update_time_s: 300.0 # Not used but kept for interface consistency
```
genevievestarke: Is there an easy way to remove this for better clarity on what's happening?

examples/hercules_input_example.yaml
```diff
@@ -0,0 +1,177 @@
+# Input YAML for Hercules simulation
```
genevievestarke: I love this, this is great!! I wonder if we could automate it in the future so it updates when we update models?

hercules/plant_components/wind_farm.py
```diff
@@ -1,43 +1,80 @@
...
+            ValueError: If wake_model is invalid or required parameters are missing.
         """
         # Store the name of this component
         self.component_name = "wind_farm"
 
-        # Store the type of this component
-        self.component_type = "Wind_MesoToPower"
+        # Infer wake_model from component_type if not provided (backward compatibility)
+        if wake_model is None:
+            wake_model = self._infer_wake_model_from_component_type(h_dict)
```
misi9170: I don't think this (or the `self._infer_wake_model_from_component_type` method) is needed? The only place `WindFarm` is instantiated is in `HybridPlant`, where we always pass `wake_model` directly.

hercules/plant_components/wind_farm.py
```diff
@@ -1,43 +1,80 @@
...
+
+        # Validate wake_model
+        if wake_model not in ["dynamic", "precomputed", "no_added_wakes"]:
+            raise ValueError(
+                f"wake_model must be 'dynamic', 'precomputed', or "
+                f"'no_added_wakes', got '{wake_model}'"
+            )
```
misi9170: This also feels a bit redundant, since the relevant code in `HybridPlant` will always generate a `KeyError` in the dictionary first?

hercules/plant_components/wind_farm.py
```diff
@@ -124,28 +161,274 @@ def __init__(self, h_dict):
...
         df_wi = interpolate_df(df_wi, time_steps_all)
 
-        # FLORIS PREPARATION
+        # INITIALIZE FLORIS BASED ON WAKE MODEL
+        if self.wake_model == "precomputed":
+            self._init_floris_precomputed(df_wi)
+        elif self.wake_model == "dynamic":
+            self._init_floris_dynamic(df_wi)
+        else:  # wake_model == "no_added_wakes"
+            self._init_floris_none(df_wi)
```
misi9170: `elif self.wake_model == "no_added_wakes"` seems more robust here, but not a big deal unless the user reaches in and sets `self.wake_model` to something else, which seems pretty unlikely

hercules/plant_components/wind_farm.py
```diff
@@ -124,28 +161,274 @@ def __init__(self, h_dict):
...
+        elif self.turbine_model_type == "dof1_model":
+            self.turbine_array = [
+                Turbine1dofModel(
+                    self.turbine_dict, self.dt, self.fmodel, self.wind_speeds_withwakes[t_idx]
+                )
+                for t_idx in range(self.n_turbines)
+            ]
+            self.use_vectorized_turbines = False
+        else:
+            raise Exception("Turbine model type should be either filter_model or dof1_model")
```
misi9170: This should probably be a `ValueError` rather than a generic `Exception`?

hercules/plant_components/wind_farm_scada_power.py
```diff
@@ -0,0 +1,346 @@
...
+        self.rated_turbine_power = np.percentile(df_scada[self.power_columns[0]], 99)
+
+        # Get the capacity of the farm
+        self.capacity = self.n_turbines * self.rated_turbine_power
+
+        self.logger.info(f"Inferred rated turbine power: {self.rated_turbine_power}")
+        self.logger.info(f"Inferred capacity: {self.capacity / 1e3} MW")
+
+        # Initialize the turbine array
+        self.turbine_array = TurbineFilterModelVectorizedSCADA(self.dt, self.scada_powers[0, :])
```
misi9170: Is `turbine_model_type` (or `turbine_file_name`) ignored for this type of wind model? Do we need to provide a warning or error if `turbine_model_type` is incorrectly specified?

misi9170 (changes requested):
I've left a handful of small comments throughout, but my main comment is on the architecture of the `WindFarm` class and its options ("dynamic", "precomputed", and "no_added_wakes"). I have a couple of concerns on this:

As far as I can tell, throughout the rest of Hercules, the "component_type" field of the input file specifies exactly which class to instantiate to model that component. The `WindFarm` class is different: the user provides a string that refers to a deprecated class, which is then mapped to a different string ("dynamic", "precomputed", or "no_added_wakes"), and a `WindFarm` object is instantiated. To me, it would be more elegant if `"component_type"` was simply `WindFarm` to match the usage in other components, and then there was a separate field `"wake_model"` (or `"wake_model_type"`, or similar) on the input file that the user can set to "precomputed", etc.

Alternatively, I think that each of these "wake_model"s on `WindFarm` could be a class on it's own. Each has it's own instantiation method, and each has a fairly different `step` method. To me, an alternative would be to have an (abstract) `WindFarm` class that all three subclasses inherit from. Each subclass would then have a method like `step_inner` that the `step` method (defined on the abstract `WindFarm`) calls. This is chaining inheritance, and has it's downsides, but to me this feels like a case where an abstract parent could be appropriate. This is a stylistic preference though, so I'm not hung up on it. If this approach is taken, then the `"component_type"` on the input file would be listed as whatever the subclass name is (e.g. `"WindFarmPrecomputedWakes"), which would also be consistent with how `"component_type"` is used throughout Hercules.

hercules/plant_components/wind_farm.py
misi9170: Leaving as is, not critical.

hercules/plant_components/wind_farm.py
```diff
@@ -1,43 +1,80 @@
...
+       Use when turbines may have individual setpoints or non-uniform operation.
 
-        This model focuses on meso-scale wind phenomena by applying a separate wind speed
-        time signal to each turbine model derived from data. It combines FLORIS wake
-        modeling with detailed turbine dynamics for wind farm performance analysis.
+    2. **precomputed**: Pre-computed FLORIS wake deficits for all conditions.
+       Use when all turbines operate uniformly (all on, all off, or uniform curtailment).
+       More efficient but less flexible than dynamic.
+
+    3. **none**: No wake modeling - wind speeds are used directly.
```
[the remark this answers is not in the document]
misi9170: Fixed.

hercules/plant_components/wind_farm.py
misi9170: Removed method

hercules/plant_components/wind_farm.py
misi9170: Removed code from `HybridPlant` and left this as is.

hercules/plant_components/wind_farm_scada_power.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # Initialize the previous powers for all turbines
+        self.prev_powers = initial_scada_powers.copy()
+
+    def step(self, scada_powers, power_setpoints):
+        """Simulate a single time step for all wind turbines simultaneously.
+
+        This method calculates the power output of all wind turbines based on the
+        given wind speeds and power setpoints. The power outputs are
+        smoothed using an exponential moving average to simulate the turbines'
```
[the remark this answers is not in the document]
genevievestarke: Renamed class to "Update" instead of "Filter"

examples/02c_wind_farm_realistic_inflow_direct/hercules_input.yaml
misi9170: This is now done.

hercules/plant_components/wind_farm_scada_power.py
misi9170: I have now added a warning that the turbine model here does not apply filtering. Plan to update at a later date.

misi9170 (approved):

PR #174, opened 2025-11-05
[BUGFIX] Hercules input yaml loader

misi9170:
#169 reconfigured the code to allow the input Hercules yaml to be loaded directly by passing the yaml filename to the `HerculesModel` constructor. However, this essentially created two copies of the same code: the `load_hercules_input()` on hercules/utilities.py and the `_load_hercules_input()` method on `HerculesModel`. Moreover, both of these methods assign new keys `"starttime"` and `"endtime"` to the `h_dict`, which is the desired behavior for the hidden method on `HerculesModel` but _not_ for the utility.

This PR fixes the situation by:
- Moving code from `HerculesModel._load_hercules_input()`, which was slightly enhanced, to `load_hercules_input()`.
- Removing the code that sets `"starttime"` and `"endtime"` from `load_hercules_input()`
- Refactoring `HerculesModel._load_hercules_input()` so that it first calls `load_hercules_input()` and then adds in the `"starttime"` and `"endtime"` keys, which appear to be needed later on
- Updating tests so that everything works as before

As well as consolidating the code, this now allows users to use load the `h_dict` directly using `h_dict = load_hercules_input("input_filename.yaml")`, modify the `h_dict` as needed, and then pass the `h_dict` to the `HerculesModel` constructor as intended; before the bug fix, this didn't work because the first call to `load_hercules_input()` would add `"starttime"` and `"endtime"` to the `h_dict`, which then triggered an error because the second call to `load_hercules_input()` disallows those keys.

### DIFF
--- a/hercules/hercules_model.py
+++ b/hercules/hercules_model.py
@@ -42,7 +42,7 @@ class HerculesModel:
         self.logger = self._setup_logging()
 
         # Load and validate the input file
-        h_dict = load_hercules_input(input_file)
+        h_dict = self._load_hercules_input(input_file)
 
         # Initialize the flattened h_dict
         self.h_dict_flat = {}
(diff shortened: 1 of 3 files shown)


paulf81 (approved):
Looks great, nice and clear, thanks @misi9170

PR #176, opened 2025-11-07
Add external data logging selection

paulf81:
This Pull Request adds the ability to select which external data channels to log (previously only logging all allowed).  Since its possible the external_data_file includes channels that are time shifts of other channels to facilitate control this option avoids bloat in the logged output.

Backwards compatiblity (with deprecation is warning) is included, such that if the log_channels are not identified then the current behavior of logging all channels is applied.

### DIFF
--- a/hercules/hercules_model.py
+++ b/hercules/hercules_model.py
@@ -204,7 +204,8 @@ class HerculesModel:
         else:
             raise ValueError(
                 f"Unsupported file format for '{filename}'. "
-                "Supported formats: CSV (.csv), Feather (.ftr, .f, .feather), Pickle (.p, .pkl, .pickle)"
+                "Supported formats: CSV (.csv), Feather (.ftr, .f, .feather), "
+                "Pickle (.p, .pkl, .pickle)"
             )
         if "time_utc" not in df_ext.columns:
             raise ValueError("External data file must have a 'time_utc' column")


docs/h_dict.md
```diff
@@ -82,3 +83,18 @@ The `h_dict` is a Python dictionary that contains all the configurations for eac
 | `costs` | dict | Cost parameters |
 | `cell_params` | dict | Cell parameters |
 | `degradation` | dict | Degradation parameters |
+
+### External Data (`external_data`)
+| Key | Type | Description | Default |
+|-----|------|-------------|---------|
+| `external_data_file` | str | Path to CSV file with external time series data | Required |
```
misi9170: Is this Required? Some simulations don't need external data files? Perhaps I'm not fully understanding

hercules/hercules_model.py
```diff
@@ -170,26 +172,40 @@ def _load_hercules_input(self, filename):
...
+        if filename_lower.endswith(".csv"):
+            df_ext = pd.read_csv(filename)
+        elif filename_lower.endswith((".feather", ".ftr")):
+            df_ext = pd.read_feather(filename)
+        elif filename_lower.endswith((".pickle", ".p", ".pkl")):
+            df_ext = pd.read_pickle(filename)
+        else:
+            raise ValueError(
+                f"Unsupported file format for '{filename}'. "
+                "Supported formats: CSV (.csv), Feather (.feather), Pickle (.pkl, .pickle)"
```
misi9170: add .ftr, .p here? We seem to mostly used .ftr, so this message seems confusing since it makes it look like .ftr is not supported

misi9170:
Note that also included in this PR is a reversion to a format of `lmp_da`, `lmp_da_xx`, and `lmp_rt` rather than `DA_LMP`, `DA_LMP_xx`, and `LMP_RT`, respectively, when referring to LMPs. The new (all caps) format was introduced in #165 but departed from previous conventions, so I've reverted to the previous convention here.

examples/05_wind_and_storage_with_lmp/hercules_runscript.py
```diff
@@ -0,0 +1,107 @@
...
+        {
+            "time_utc": time_points,
+            "lmp_da": np.full(49, 10.0),  # Constant $10
+            "lmp_rt": np.linspace(0, 50, 49),  # Ramp from $0 to $50
+        }
+    )
+
+    # Save to CSV
+    df.to_csv(lmp_data_path, index=False)
+    print(f"Created {lmp_data_path}")
```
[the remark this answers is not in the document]
paulf81: sure thing

hercules/hercules_model.py
paulf81: good catch

docs/h_dict.md
misi9170: Ah, interesting. So, having the top-level `external_data` field with nothing in it is not allowed, is that correct? I wonder if there is a protocol here; it seems that there might be cases when it's nice to be able to have "empty" top levels (say, an empty `wind_farm`), just as placeholders; but if I'm understanding correctly, that is not currently supported. In that case, I see your point, and I suppose this is fine (Required if `external_data` is provided at all)

PR #181, opened 2025-11-14
Clean up

paulf81:
This is a small pull request just to clean out some linting and formatting that I noticed would get pulled into other PRs when I run a global ruff check/format and so just wanted to collect these into one PR and combine with:

- Quick scan for typos and formatting erros
- Small doc improvements


paulf81:
ok @misi9170 I added a CI check that formatting is correct (and confirmed CI failed when I intentionally left some bad formatting in).  Now it should pass.  Also I added the missing variables back to docs/docstring/default of NSRDB download so good for review.  Thanks!

PR #189, opened 2025-12-18
Open cycle gas turbine model and thermal power plant base class

paulf81:
This PR adds a base class for thermal components, including both gas and coal plants.  The modeling is based largely on 3 papers:

[1] Agora Energiewende (2017): Flexibility in thermal power plants
     With a focus on existing coal-fired power plants.
[2] "Impact of Detailed Parameter Modeling of Open-Cycle Gas Turbines on
    Production Cost Simulation", NREL/CP-6A40-87554, National Renewable
    Energy Laboratory, 2024.
[3] Deane, J.P., G. Drayton, and B.P. Ó Gallachóir. "The Impact of Sub-Hourly
    Modelling in Power Systems with Significant Levels of Renewable Generation."
     Applied Energy 113 (January 2014): 152–58.
     https://doi.org/10.1016/j.apenergy.2013.07.027.
[4] IRENA (2019), Innovation landscape brief: Flexibility in conventional power plants,
    International Renewable Energy Agency, Abu Dhabi.
[5] M. Oakes, M. Turner, " Cost and Performance Baseline for Fossil Energy Plants, Volume 5:
    Natural Gas Electricity Generating Units for Flexible Operation," National Energy
    Technology Laboratory, Pittsburgh, May 5, 2023.
[6] I. Staffell, "The Energy and Fuel Data Sheet," University of Birmingham, March 2011.
    https://claverton-energy.com/cms4/wp-content/uploads/2012/08/the_energy_and_fuel_data_sheet.pdf

In this PR, in addition to the base class: `ThermalComponentBase` a first subclass is added: `OpenCycleGasTurbine`, implementing an open-cycle (CT, peaker) NG component.  The PR includes an example demonstrating `OpenCycleGasTurbine`, docs for the new code and tests for both.

@misi9170 and I chose to use efficiency, as in [4,5] over heat rate, as being a more general concept.  Further, SI units like J and m3 are used for consistency over BTU, etc in earlier versions of this PR

Still todo:

- [ ] @genevievestarke could you review this model and compare with what is in HOPP/H2I?


hercules/plant_components/combustion_turbine_simple.py
```diff
@@ -0,0 +1,401 @@
...
+        """Calculate fuel consumed based on power output and heat rate.
+
+        Args:
+            power_output (float): Current power output in kW.
+
+        Returns:
+            float: Fuel consumed this timestep in kJ.
+        """
+
+        # TODO: Is this correct?  Should we be getting to m^3 of natural gas?
```
paulf81: @genevievestarke check this one?

tests/combustion_turbine_simple_test.py
```diff
@@ -0,0 +1,13 @@
+import copy
+
+from hercules.plant_components.combustion_turbine_simple import CombustionTurbineSimple
+
+from .test_inputs.h_dict import (
+    h_dict_combustion_turbine,
+)
+
+
```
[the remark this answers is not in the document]
paulf81: This was just a placeholder

paulf81:
ok @genevievestarke  and @misi9170 , I now wrote a set of tests and updated the docs so this one should be ready for review.

paulf81:
ok @genevievestarke  and @misi9170 , this PR is now fully reworked following our discussion.  It most closely follows the model descriptions in:

[1] Agora Energiewende (2017): Flexibility in thermal power plants
     With a focus on existing coal-fired power plants.
     
I redid the top comment box on this PR and added 2 todo items if you don't mind to have a look, or can just talk with me about it.
     
@brookeslawski  I added you as a reviewer in case you have a few minutes would value your feedback, but understand if you don't!

hercules/hybrid_plant.py
```diff
@@ -116,7 +117,10 @@ def get_plant_component(self, component_name, h_dict):
         if component_type == "ElectrolyzerPlant":
             return ElectrolyzerPlant(h_dict)
 
-        raise Exception("Unknown component_type: ", component_type)
+        if component_type == "OpenCycleGasTurbine":
+            return OpenCycleGasTurbine(h_dict)
+
+        raise Exception("Unknown component_type: ", h_dict[component_name]["component_type"])
```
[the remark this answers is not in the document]
paulf81: works for me

examples/07_ocgt/hercules_input.yaml
```diff
@@ -0,0 +1,44 @@
...
+  interconnect_limit: 100000  # kW (100 MW)
+
+open_cycle_gas_turbine:
+  component_type: OpenCycleGasTurbine
+  rated_capacity: 100000  # kW (100 MW)
+  min_stable_load_fraction: 0.2  # 20% minimum operating point
+  ramp_rate_fraction: 0.1  # 10%/min ramp rate
+  run_up_rate_fraction: 0.05  # 5%/min run up rate
+  hot_startup_time: 490.0  # 7 minutes
+  cold_startup_time: 580.0  # 8 minutes
```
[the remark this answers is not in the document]
paulf81: Fixing

examples/07_ocgt/hercules_runscript.py
```diff
@@ -0,0 +1,80 @@
...
+
+        """
+        current_time = h_dict["time"]
+
+        # Determine power setpoint based on time
+        if current_time < 40 * 60:  # 40 minutes in seconds
+            # Before 40 minutes: keep turbine off
+            power_setpoint = 0.0
+        elif current_time < 120 * 60:  # 120 minutes in seconds
+            # Between 60 and 120 minutes: run at full capacity
```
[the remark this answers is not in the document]
paulf81: Clarified

examples/07_ocgt/plot_outputs.py
```diff
@@ -0,0 +1,84 @@
+# Plot the outputs of the simulation for the NGCT example
```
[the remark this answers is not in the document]
paulf81: Fixed

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,436 @@
...
+        # ====================================================================
+        elif self.state_num == self.STATE_ON:
+            # Check if we can shut down (min_up_time satisfied)
+            can_shutdown = self.time_in_state >= self.min_up_time
+
+            if power_setpoint <= 0 and can_shutdown:
+                # Transition to shutdown sequence
+                self.state_num = self.STATE_STOPPING
+                self.time_in_state = 0.0
+
```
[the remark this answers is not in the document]
paulf81: Agree with this suggestion, implementing the change

hercules/utilities.py
```diff
@@ -27,21 +27,23 @@ def get_available_component_names():
...
     ]
 
 
 def get_available_generator_names():
     """Return available generator component names.
 
-    Returns power generators (wind_farm, solar_farm), excluding storage and conversion
-    components.
+    Returns power generators (wind_farm, solar_farm, combustion_turbine), excluding
+    storage and conversion components.
```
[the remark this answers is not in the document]
paulf81: fixed

tests/open_cycle_gas_turbine_test.py
```diff
@@ -0,0 +1,93 @@
...
+import pytest
+from hercules.plant_components.open_cycle_gas_turbine import OpenCycleGasTurbine
+
+from .test_inputs.h_dict import (
+    h_dict_open_cycle_gas_turbine,
+)
+
+
+def test_init_from_dict():
+    """Test that ThermalComponentBase can be initialized from a dictionary."""
```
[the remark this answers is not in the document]
paulf81: fixed

tests/thermal_component_base_test.py
```diff
@@ -0,0 +1,394 @@
...
+def test_init_from_dict():
+    """Test that ThermalComponentBase can be initialized from a dictionary."""
+    tpb = ThermalComponentBase(copy.deepcopy(h_dict_thermal_component))
+    assert tpb is not None
+
+
+def test_invalid_inputs():
+    """Test that ThermalComponentBase raises an error for invalid inputs."""
+    h_dict = copy.deepcopy(h_dict_thermal_component)
+
```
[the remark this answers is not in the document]
paulf81: Removed

hercules/plant_components/combustion_turbine_simple.py
paulf81: Just updating @misi9170  and I decided to end on m3, this seems standard enough

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,436 @@
...
+            # Ramp up using run_up_rate
+            startup_power = (self.time_in_state - self.hot_readying_time) * self.run_up_rate
+
+            # Check if ramping is complete
+            if startup_power >= self.P_min:
+                self.state_num = self.STATE_ON
+                self.time_in_state = 0.0
+                return startup_power
+
+            return startup_power
```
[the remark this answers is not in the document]
paulf81: ok added this limit

examples/07_ocgt/hercules_runscript.py
```diff
@@ -0,0 +1,80 @@
...
+hmodel = HerculesModel("hercules_input.yaml")
+
+
+class ControllerNGCT:
+    """Controller implementing the NGCT schedule described in the module docstring."""
+
+    def __init__(self, h_dict):
+        """Initialize the controller.
+
+        Args:
```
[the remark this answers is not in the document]
paulf81: fixed

docs/examples/07_open_cycle_gas_turbine.md
```diff
@@ -0,0 +1,66 @@
...
+
+```bash
+python plot_outputs.py
+```
+
+The plot shows:
+- Power output over time (demonstrating ramp constraints and minimum stable load)
+- Operating state transitions
+- Fuel consumption tracking
+- Heat rate variation with load
```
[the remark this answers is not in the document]
paulf81: Outdated

paulf81:
ok @misi9170  and @genevievestarke , the latest refactor is done and the code is ready for review again

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,628 @@
...
+        if self.power_output < 0 or self.power_output > self.rated_capacity:
+            raise ValueError(
+                "initial_conditions['power'] (initial power) "
+                "must be between 0 and rated_capacity (inclusive)"
+            )
+        if self.state_num not in self.STATE_NAMES:
+            raise ValueError(
+                f"initial_conditions['state_num'] must be one of {list(self.STATE_NAMES.keys())}"
+            )
+
```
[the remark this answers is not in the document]
paulf81: This makes sense, but feels a little like adding bulk for a pretty odd edge case.  @misi9170 , do you have a preference here?

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,628 @@
...
+            # Ramp up using run_up_rate
+            startup_power = (self.time_in_state - self.hot_readying_time) * self.run_up_rate
+
+            # Check if ramping is complete
+            if startup_power >= self.P_min:
+                self.state_num = self.STATE_ON
+                self.time_in_state = 0.0
+                return startup_power
+
+            # Limit to below P_max (edge case)
```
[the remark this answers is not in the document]
paulf81: this is done

paulf81:
@misi9170 one more question for you, right now the member variable `fuel_consumption` is in units m3, while `fuel_consumption_kg` is in kg.  Should I rename   `fuel_consumption` to `fuel_consumption_m3` so there is no ambiguity?

hercules/plant_components/thermal_component_base.py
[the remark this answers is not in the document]
misi9170: @paulf81 I think this could lead to confusion, honestly. Perhaps the initial `state_num` should _not_ be user-specifiable; that is, it is set by whatever the user supplies as the initial `power`. That will be clear for powers above minimum operation (state ON) and powers at or below 0 (state OFF). It will be less clear if they specify a power somewhere in between, and perhaps we should just through a useful error if the initial `power` is set to a transition value (that is, the system must start either ON or OFF, as specified by the initial `power`).

This way, as well, users may not have to deal themselves with the state numbering system.

misi9170:
> @misi9170 one more question for you, right now the member variable `fuel_consumption` is in units m3, while `fuel_consumption_kg` is in kg. Should I rename `fuel_consumption` to `fuel_consumption_m3` so there is no ambiguity?

Perhaps. Another option might be to rename the member variables `fuel_volume_consumption` and `fuel_mass_consumption`, with units of m^3 and kg, respectively.

misi9170:
> @misi9170 and I chose to use efficiency, as in [4,5] over heat rate, as being a more general concept. Further, SI units like J and m3 are used for consistency over BTU, etc in earlier versions of this PR

Sounds good to me, thanks

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,628 @@
...
+    STATE_HOT_STARTING = 1
+    STATE_WARM_STARTING = 2
+    STATE_COLD_STARTING = 3
+    STATE_ON = 4
+    STATE_STOPPING = 5
+
+    # Time constants
+    #       Note the time definitions for cold versus warm versus hot starting are hard
+    #   coded and based on the values in [5].
+    HOT_START_TIME = 8 * 60 * 60  # 8 hours
```
genevievestarke: Do we set the cold start time, too?

genevievestarke (commented):
I think this is looking really great!
My top comment is if we want to enable setting the 'time_in_state` variable to something other than zero at initialization in case we need to start or stop the model in less than an hour (if it's been on for awhile, for example, and we want to turn it off). I think currently, it won't do anything until it gets beyond the minimum time in state, which may be confusing.

genevievestarke:
Ok, I've compared it to the H2I NG model. That model calculates electricity by using electricity = natural_gas / heat_rate_mmbtu_per_mw. The model uses MMBTUs rather than what Hercules has. The model constrains the production based on natural gas feedstock availability, but not dynamics. It also has financial metrics, like cost of the natural gas, etc. It seems like these two models will be complementary, though perhaps we should think about aligning units (or writing down some automatic unit converters).

hercules/plant_components/thermal_component_base.py
paulf81: Ah, this could've been clearer, HOT_START is less than HOT_START_TIME, warm stat less than WARM_START_TIME and then cold start is if greater than both numbers.  Added some clarifying comments

hercules/plant_components/thermal_component_base.py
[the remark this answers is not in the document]
paulf81: Ok, code updated along the lines charted here and our discussion.  `self.state` is set via `enum` as is done in FLORIS.  Then also, `state` is not assigned directly by initial condition, but instead inferred from the initial value of `power`.  Finally, `time_in_state` is always initialized to the appropriate `min_time` so it is ready to transition when the simulation begins.

paulf81:
> I think this is looking really great! My top comment is if we want to enable setting the 'time_in_state` variable to something other than zero at initialization in case we need to start or stop the model in less than an hour (if it's been on for awhile, for example, and we want to turn it off). I think currently, it won't do anything until it gets beyond the minimum time in state, which may be confusing.

This was a good catch.  I changed the code so that time_in_state is automatically initialized to the appropriate `min_time_...` value so it always ready to transition at the start of a simulation.  @misi9170  and I thought this saves too many options in the API.  And in cases where it's important to set the initial `time_in_condition` to something else, this can be done programatically

paulf81:
> Ok, I've compared it to the H2I NG model. That model calculates electricity by using electricity = natural_gas / heat_rate_mmbtu_per_mw. The model uses MMBTUs rather than what Hercules has. The model constrains the production based on natural gas feedstock availability, but not dynamics. It also has financial metrics, like cost of the natural gas, etc. It seems like these two models will be complementary, though perhaps we should think about aligning units (or writing down some automatic unit converters).

This is good to hear @genevievestarke .  An earlier version of this code used the mixed unit heat_rate value as well but @misi9170  and I discussed it and thought having all units in SI within Hercules the most consistent.  I think it's a pretty simple conversion between % efficiency when we compare across codes

paulf81:
> > @misi9170 one more question for you, right now the member variable `fuel_consumption` is in units m3, while `fuel_consumption_kg` is in kg. Should I rename `fuel_consumption` to `fuel_consumption_m3` so there is no ambiguity?
> 
> Perhaps. Another option might be to rename the member variables `fuel_volume_consumption` and `fuel_mass_consumption`, with units of m^3 and kg, respectively.

That's a good suggestion, implemented

docs/examples/07_open_cycle_gas_turbine.md
misi9170, line 1: Consider naming this, as well as the actual example directory, open_cycle_gas_turbine for clarity

docs/_static/thermal_startup_ramp.svg
misi9170, line 1: What is the _static/ folder for? Not something I'm familiar with

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+        power_fraction = power_output / self.rated_capacity
+
+        # Interpolate efficiency (numpy.interp clamps to endpoints by default)
+        efficiency = np.interp(
+            power_fraction, self.efficiency_power_fraction, self.efficiency_values
+        )
+
+        return efficiency
+
+    def _calc_fuel_consumption(self, power_output):
```
misi9170: Is there any chance a child class might want to overload this method? If so, I would remove the leading underscore to "unhide" it. Not sure if that might also apply to other methods on this base class

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+
+        # Calculate current efficiency
+        efficiency = self._calc_efficiency(power_output)
+
+        # Calculate fuel consumption
+        # fuel_volume (m³) = power (W) * dt (s) / (efficiency * hhv (J/m³))
+        # Convert power from kW to W (multiply by 1000)
+        fuel_m3 = (power_output * 1000.0) * self.dt / (efficiency * self.hhv)
+
+        return fuel_m3
```
misi9170: I'm thinking that perhaps this ought to be a consumption rate, unit m3/s, to be more consistent with channels in Hercules (power, hydrogen production in kg/s, etc).

tests/test_inputs/h_dict.py
```diff
@@ -96,6 +96,61 @@
...
+    "warm_startup_time": 120.0,  # s (must be >= ramp_time of 60s)
+    "cold_startup_time": 120.0,  # s (must be >= ramp_time of 60s)
+    "min_up_time": 10.0,  # s
+    "min_down_time": 10.0,  # s
+    "log_channels": [
+        "power",
+        "state",
+        "efficiency",
+        "fuel_volume_consumption",
+        "fuel_mass_consumption",
```
misi9170: Consider making these rates, `fuel_volume_consumption_rate`, `fuel_mass_consumption_rate`

examples/07_ocgt/plot_outputs.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Plot the fuel consumption
+ax = axarr[3]
+ax.plot(
+    time_minutes,
+    df["open_cycle_gas_turbine.fuel_volume_consumption"],
+    label="Fuel Consumption",
+    color="orange",
+)
+ax.set_ylabel("Fuel [m³/timestep]")
```
misi9170: If switching to fuel consumption rates, rather than absolute fuel consumption per time step, make sure to update this plot and axis labels. Even if the logged consumption is in absolute units, I'd recommend normalizing here by `dt` so that the units on the y axis can be [m^3/s] (I find it hard to wrap my head around "per timestep" units)

hercules/plant_components/open_cycle_gas_turbine.py
```diff
@@ -0,0 +1,130 @@
...
+        if "hot_startup_time" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["hot_startup_time"] = 420.0
+        if "warm_startup_time" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["warm_startup_time"] = 480.0
+        if "cold_startup_time" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["cold_startup_time"] = 480.0
+        if "min_up_time" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["min_up_time"] = 1800.0
+        if "min_down_time" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["min_down_time"] = 3600.0
```
misi9170, line 106: A more elegant way to set up defaults would be to specify a defaults dictionary, and then use something like
```python
default_dict = {"a": 1, "b": 2}
passed_dict = {"b": 3, "c": 4}

final_dict = {**default_dict, **passed_dict}
```

It looks like as of python 3.9, there is a slightly cleaner version:
```python
final_dict = default_dict | passed_dict
```

Even better might be to use [`attrs`](https://www.attrs.org/en/stable/), but that should then probably be done across Hercules and should be its own PR.

hercules/plant_components/open_cycle_gas_turbine.py
```diff
@@ -0,0 +1,130 @@
...
+            h_dict[self.component_name]["hhv"] = 39050000  # J/m³ (39.05 MJ/m³)
+
+        # Default fuel density for natural gas (0.768 kg/m³) from [6]
+        if "fuel_density" not in h_dict[self.component_name]:
+            h_dict[self.component_name]["fuel_density"] = 0.768  # kg/m³
+
+        # Call the base class init
+        super().__init__(h_dict)
+
+    def get_initial_conditions_and_meta_data(self, h_dict):
```
misi9170: Should this be on the base class? Or is this specific to the child?

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+
+    @property
+    def state_name(self):
+        """Return the name of the current state.
+
+        Returns:
+            str: Current state name ("off", "hot starting", "warm starting", "cold starting",
+                "on", or "stopping").
+        """
+        return self.STATE_NAMES[self.state]
```
misi9170: Not sure how necessary this is any more? Doesn't hurt, I guess

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+                    (both as fractions 0-1)
+        """
+
+        # Both the component name and type are defined in the subclass
+        # But for testing purposes, we need to set the component name here
+        # if not defined in the subclass
+        if not hasattr(self, "component_name"):
+            self.component_name = "thermal_component"
+        if not hasattr(self, "component_type"):
+            self.component_type = "ThermalComponentBase"
```
misi9170: Maybe I'm not quite following---couldn't these be passed in the test `h_dict`? I don't really know if there are rules around this, but it feels fishy to have source code that is _only_ used for testing

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+        if self.hot_startup_time < 0:
+            raise ValueError("hot_startup_time must be greater than or equal to 0")
+        if self.warm_startup_time < 0:
+            raise ValueError("warm_startup_time must be greater than or equal to 0")
+        if self.cold_startup_time < 0:
+            raise ValueError("cold_startup_time must be greater than or equal to 0")
+        if self.min_up_time < 0:
+            raise ValueError("min_up_time must be greater than or equal to 0")
+        if self.min_down_time < 0:
+            raise ValueError("min_down_time must be greater than or equal to 0")
```
misi9170, line 165: Again, maybe we should be looking at attrs....

hercules/plant_components/thermal_component_base.py
```diff
@@ -0,0 +1,635 @@
...
+
+        # Sort arrays by power_fraction for proper interpolation
+        sort_idx = np.argsort(self.efficiency_power_fraction)
+        self.efficiency_power_fraction = self.efficiency_power_fraction[sort_idx]
+        self.efficiency_values = self.efficiency_values[sort_idx]
+
+        # Initialize efficiency and fuel consumption
+        self.efficiency = self._calc_efficiency(self.power_output)
+        self.fuel_volume_consumption = 0.0  # m³/timestep
+        self.fuel_mass_consumption = 0.0  # kg/timestep
```
misi9170: rates?

hercules/plant_components/open_cycle_gas_turbine.py
misi9170: I see now that the base raises a `NotImplementedError` for this method, which I think is fine. Looking at the values being accessed in this method, though, they seem general to all thermal plants, so perhaps this could live on the base.

hercules/plant_components/thermal_component_base.py
misi9170: Extremely minor point, but given that these methods are not particularly long names, it might be more pythonic to just name them `_calculate_fuel_consumption`, etc (i.e., spell out `calculate` rather than shortening it). These are currently hidden methods, so that makes this point extremely nitpicky---you're welcome to ignore it :)

misi9170 (changes requested):
A bunch of minor comments for you to consider, @paulf81 . 

The two more far-reaching ones are:
- consider specifying fuel consumption as a rate throughout
- (likely a larger discussion) consider using attrs; there is a lot of boilerplate initialization code in here

paulf81:
> A bunch of minor comments for you to consider, @paulf81 .
> 
> The two more far-reaching ones are:
> 
> * consider specifying fuel consumption as a rate throughout
> * (likely a larger discussion) consider using attrs; there is a lot of boilerplate initialization code in here

Created #203 to track the attrs suggestion

Can do on specifying fuel consumption as a rate

paulf81:
@misi9170 flagging in the course of setting fuel consumption over to a rate, realized that I had mistakenly combined HHV as the energy value of the gas, with a table giving net efficiency with respect to LHV.  I've fixed this, and in the course of fixing it decided to add default values based on one of the reference papers, and also adding a lot of comments that efficiency is respect to HHV always.  My understanding is you can use either as long as you're consistent, but feel free to disagree @genevievestarke  and @misi9170 !

docs/_static/thermal_startup_ramp.svg
paulf81: I understand this is the conventional place to put images when building docs with jupyter-book

docs/examples/07_open_cycle_gas_turbine.md
paulf81: Renaming the example now

hercules/plant_components/open_cycle_gas_turbine.py
paulf81: Agreed, I'll move to base class and we can reconsider down the road but for now this seems right

hercules/plant_components/open_cycle_gas_turbine.py
paulf81: Yeah, maybe this can be bundled under the new attrs issue [#203](https://github.com/NatLabRockies/hercules/issues/203)

hercules/plant_components/thermal_component_base.py
paulf81: Yeah, I saw that, let's remove for simplicity's sake

hercules/plant_components/thermal_component_base.py
paulf81: Yeah, I did this while spinning this all up but let me see if I can now remove it, it does feel non-standard

paulf81: ok, the more pythonic thing would be to make these class attributes, then they automatically act as defaults unless over-ridden by subclass, making this change now

paulf81: but noting this could also be rolled up to changes in the attrs upgrade in #203

hercules/plant_components/thermal_component_base.py
paulf81: should be fixed now

tests/test_inputs/h_dict.py
paulf81: should be done

hercules/plant_components/thermal_component_base.py
paulf81: done

hercules/plant_components/thermal_component_base.py
paulf81: Done!

paulf81:
Thanks for comments @misi9170 , think I've now made all requested changes

genevievestarke (approved):
I think this looks good to me!
One question: Is this only for aerodrive turbines? There are apparently aerodrive and heavy-frame turbines in the open cycle category. The heavy-frame seem to be more suited for baseload power.

paulf81:
> I think this looks good to me! One question: Is this only for aerodrive turbines? There are apparently aerodrive and heavy-frame turbines in the open cycle category. The heavy-frame seem to be more suited for baseload power.

A good question, I'm not sure the answer.  I was thinking of the split as between combined-cycle and open-cycle, and something like, most open-cycle plants will be aeroderivative because they're focused on peaking, while most combined-cycle will use frame as more optimized toward running more often.  There's exceptions I think but maybe we say yes, these parameters assume an aeroderivative turbine.  Since the parameters come from the references I can also check what they assume, one second

Update: Confirmed the OC turbines in [5] are explicitly aeroderivative

PR #199, opened 2026-01-15
Update URLs

misi9170:
Update URLS: 
- github.com/NREL/ -> github.com/NatLabRockies/
- nrel.github.io -> natlabrockies.github.io

Prior to merger, check final location of WETOSoftware documentation page

@genevievestarke : Would you rather merge this straight into main without a new release (and then merge back into develop), or would you prefer to merge into develop and then create a patch release to get it onto main? I'm planning on the latter for FLORIS, but doing it slightly more informally on Hercules is probably fine if you'd rather not jump through the hoops of releasing Hercules at the moment.

This PR should _not_ be merged until the organization name actually changes to github.com/NatLabRockies

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -6,6 +6,6 @@ authors:
     given-names:
 title: "HERCULES"
 version: 2
-url: https://github.com/NREL/hercules
+url: https://github.com/NatLabRockies/hercules
 
 date-released: 2025-12-31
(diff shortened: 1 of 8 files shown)


genevievestarke:
That makes sense! I think we can see where we're at when the name changes. If we're ready for an official release, I can go ahead and release it, but otherwise we'll do it informally.

paulf81:
Thanks!!

PR #204, opened 2026-02-19
Minor clean up

paulf81:
This minor clean up pull requests applies 3 small changes:

1) Lines where we had previously omitted to apply `dtype=hercules_float_type` are corrected throughout the code
2) A few tests where this change causes the test to fail on numerical difference are corrected, by either declaring the array within the test code to also be `hercules_float_type` or by using an approximate comparison
3) Two test files were found to have names refernecing the old wind farm class so these are corrected to avoid confusion going forward.  Specifically:
tests/wind_meso_to_power_test.py → tests/wind_farm_test.py
tests/wind_meso_to_power_precom_floris_test.py → tests/wind_farm_precom_floris_test.py
With some of the test function names and documentation likewise corrected.


paulf81:
@genevievestarke based on the CI failure seems like this update to electrolyzer is timely!

tests/wind_farm_dynamic_floris_test.py
genevievestarke, line 1: For this file, we're assuming that the dynamic mode is the default mode. Should we name it `wind_farm_dynamic_floris_test` for clarity?

paulf81: Good call!  making the change

genevievestarke (approved):

PR #205, opened 2026-02-19
Update python versions

paulf81:
Currently in Hercules the minimum python version is specified as 3.9, and then in CI we test 3.9, 3.10 and 3.11.  This PR makes a few small changes:

1) Updates minimum version to 3.10
2) Drops CI testing of 3.9 and adds testing of 3.12 and 3.13

The reasons for (1) are:
1) Numpy 2.1+ no longer supports 3.9 and Hercules has a ~2.0 dependency to numpy which would include 2.1
2) Python 3.9 is end of life: https://devguide.python.org/versions/

The reasons for (2) are that we were implicitly supporting these versions without testing

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: False
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
+        python-version: [ "3.11", "3.12", "3.13"]
         os: [ubuntu-latest] #, macos-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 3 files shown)


misi9170:
Any reason not to add python 3.14 also, since that is the "current" version?

paulf81:
> Any reason not to add python 3.14 also, since that is the "current" version?

You're right, this was an oversight on my part, added...

paulf81:
Noting that I removed:
3.10 from testing because it isn't supported by electrolyzer
3.14 because it isn't supported by pysam

Can revisit in the future and both versions are technically "allowed" by `pyproject.toml`.  3.10 will work if you don't pip install the electrolyzer module, but 3.14 will not yet work but will probably in the future.

@misi9170 @genevievestarke

misi9170 (approved):

paulf81:
thanks @misi9170 !

genevievestarke (approved):

PR #214, opened 2026-02-24
[DOCS] Update installation instructions

misi9170:
The [existing installation instructions](https://natlabrockies.github.io/hercules/install.html) incorrectly refer to `git switch`ing to the `v2` branch. That branch no longer exists, and moreover, `v2` is a tag that can't be `switch`ed to but can be `checkout`ed, which creates confusion.

This PR updates the installation instructions. In doing so, I also added a missing `pip install` command for the electrolyzer repo and clarified that `develop` need only be switched to if the user plans to actually perform code development.

Thanks @jfrederik-nrel for identifying this issue!

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -36,7 +36,7 @@ conda activate hercules
 Alternatively, you can use uv to create a new environment for hercules.  This will create a new environment in the current directory.
 
 ```bash
-uv venv
+uv venv --python 3.13
 source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


paulf81:
This looks good @misi9170 .  I think for consistency we should change the uv install command to:
`uv venv --python 3.13`

paulf81 (approved):

PR #217, opened 2026-02-27
PyPI publishing workflow

misi9170:
Adds a workflow to publish Hercules as a package on PyPI. Hercules would first appear on PyPI after this pull request arrives on the main branch and a new release for Hercules is created.

Set up using PyPI's Trusted Publishing workflow.

Unforunately, [`hercules`](https://pypi.org/project/hercules/) is taken as a PyPI package name (on 0.0, with no releases since 2014.... oh well). We could alternatively publish under the following, which appear to be available:
- [`herc`](https://pypi.org/project/herc/) 
- [`nlr-hercules`](https://pypi.org/project/nlr-hercules/)

Any other suggestions?

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,34 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# Published via GitHub Actions as a PyPI Trusted Publisher.
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+# and: https://docs.pypi.org/trusted-publishers/
+
+name: Upload Python Package
+
+on:
+  release:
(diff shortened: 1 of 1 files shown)


genevievestarke:
I think I would prefer nlr-hercules. And that seems consistent with naming conventions that have been used in the past!

paulf81:
> I think I would prefer nlr-hercules. And that seems consistent with naming conventions that have been used in the past!

Agree, this is better option unless we rename repo to herc

misi9170:
Planning for `nlr-hercules` and I have PyPI primed to receive the package under that name. Going to delay merging this until early next week so we can decide if we want to merge this PR into develop and release a version 2.1 with various changes; or if we want to merge straight to main and release a version 2.0.1, with larger changes following in a subsequent v2.1

Issue #218, opened 2026-03-02
Latest version of NREL-Pysam not currently supported


misi9170:
Maybe dependabot would have told us this on the next release, but [NREL-Pysam v7 was released in April 2025](https://github.com/NatLabRockies/pysam/releases) and Hercules currently requires v6. I've tried running with v7, and unfortunately there is an incompatibility when running example 03_wind_and_solar. I haven't checked other examples.

I'll look into what would need to change on the Hercules side to support v7, and if it's possible to simultaneously support both v6 and v7 of `nrel-pysam`.

PR #219, opened 2026-03-02
Enable compatibility with `NREL-PySAM` v6 and v7

misi9170:
Currently, Hercules [requires `NREL-PySAM v6`](https://github.com/NatLabRockies/hercules/blob/efb5fd666d7f1b66329177abe779c00c993ec7df/pyproject.toml#L32). However, [v7 was released in April 2025](https://github.com/NatLabRockies/pysam/releases/tag/v7.0.0) and the [latest version is v7.1.0](https://github.com/NatLabRockies/pysam/releases/tag/v7.1.0) (released August 2025).

Hercules, as is, is not compatible with v7 of `NREL-PySAM`. This "fixes" this by adding a couple of `try`/`except` clauses on a few lines that generate errors. *However, I am not sure if this is a legitimate fix or if I am changing the behavior*.

With the update:
- tests pass using both v6.0.1 and v7.1.0
- example 03_wind_and_solar runs to completion and produces visually the same result:

Using v6.0.1:
image

Using v7.1.0:
image

Assuming we no longer need to use the lines that are `except`ed in this change, we could simply require v7. However, this is a change that will affect users directly, as they'll need to update their local environments to keep using Hercules. The change I've implemented here allows users to stay on v6, or to move up to v7.

*If a valid correction*, addresses #218

### DIFF
--- a/hercules/plant_components/solar_pysam_pvwatts.py
+++ b/hercules/plant_components/solar_pysam_pvwatts.py
@@ -68,10 +68,6 @@ class SolarPySAMPVWatts(SolarPySAMBase):
         system_model.assign(self.model_params)
 
         system_model.AdjustmentFactors.adjust_constant = 0
-        try:
-            system_model.AdjustmentFactors.dc_adjust_constant = 0
-        except Exception:
-            pass
 
         # Save the system model
         self.system_model = system_model
@@ -98,10 +94,6 @@ class SolarPySAMPVWatts(SolarPySAMBase):
 
         # Assign the full solar resource data
         self.system_model.SolarResource.assign({"solar_resource_data": solar_resource_data})
-        try:
-            self.system_model.AdjustmentFactors.assign({"constant": 0})
-        except Exception:
-            pass
 
         # Execute the model once for all time steps
         self.system_model.execute()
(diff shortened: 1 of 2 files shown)


paulf81:
I think that could work, I think I added those lines when tinkering with this module and not sure how critical.  Maybe @brookeslawski would you mind to give this change a look over?

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -68,7 +68,10 @@ def _create_system_model(self):
         system_model.assign(self.model_params)
 
         system_model.AdjustmentFactors.adjust_constant = 0
-        system_model.AdjustmentFactors.dc_adjust_constant = 0
+        try:
```
elenya-grant: if you want to keep backward compatibility with PySAM v6 you could do:
```python
import PySAM

if int(PySAM.__version__.split(".")[0]) < 7:
   system_model.AdjustmentFactors.dc_adjust_constant = 0
```

instead of the try-except

misi9170: Going to see whether running with v6 with these lines just removed still passes tests

misi9170:
Ok, I've now simply removed the two lines that caused issues. Tests still pass, and example 03 still provides (visually) the same output, so perhaps that's good enough?

PR #216, opened 2026-02-26
Reorganize component name and type

paulf81:
This pull request is meant to address Issue #207.  The main points are to:

1) Clarify the usage and purpose of `component_name` and `component_type`
2) Facilitate allowing multiple instances of the same type of component
3) Maintain backward compatibility

This PR accomplished this via:

1) Changing `component_name` to be a unique identifier that can be any string.  It's backwards-compatible since it is still defined as the YAML top level key and the names for example in the examples like `wind_farm` still work since there is only one thing called `wind_farm` in those examples
2) Keeping `component_type` to represent the specific class use to define the unit, however making that more explicit by automatically defining the type in the `ComponentBase` using the `__name__` method
3) Adding a `component_category` that captures what the `component_name` used to represent.  These are class attributes defined at least in each base class and then optionally in subclasses

Then also the following organization steps are taken:

- [x] All components and base classes updated around this new pattern
- [x] Components now require a `component_name` to be passed in.  This is a smallish change handled mainly by `HybridPlant` but all tests of the components themselves need this added.
- [x] `HybridPlant` updated to use these more explicit variables
- [x] `utilities.py` simplified since many of the types/classes are kept within `HybridPlant`
- [x] A few additional tests added for the new features
- [x] Documentation updated

Issue #210, opened 2026-02-23
More tightly modularize resources/ folder


paulf81:
Currently in resource folder is:
 - `hercules/resource/upsample_wind_data.py`
 - `hercules/resource/wind_solar_resource_downloader.py`

This issue captures the intention that these elements could be broken down into clearer modules (collecting maybe some of the other sub-issues of this parent issue).  One idea:

 - A module file for downloads of particular points
 - A module file for downloads of areas with resolution
 - A module for conversion into Hercules format
 - A module for upsampling in time and space

paulf81:
#246 is now the first step in this process

PR #221, opened 2026-03-02
Functionalize resource module and add tests

paulf81:
This draft PR is a first step towards addressed #210 .  Rather then doing that all in one PR, thought a nicer thing is a PR that simply reorganizes what is already there in `hercules/resource/wind_solar_resource_downloader.py` into smaller functions and add tests for those functions.  Currently, the main means of checking the resourece modules are the examples in `examples/inputs`, but if we chunk smaller operations into smaller functions, then all but the download itself can be tested with synthetic data.  This can all be done without any changes that say effect the files in `examples/inputs`.  

Note: this will look like more changes than it is until #216 is merged

Steps:

- [ ] Break large functions in `hercules/resource/wind_solar_resource_downloader.py` into smaller functions, and those that can be applied to several relocate to a new file `hercules/resource/resource_utilities.py`
- [ ] Iterate to happy we've hit a small enough level (can also consider rewriting some aspects but thought keeping scope to simply re-arranging a nice start)
- [ ] Add tests for functions in `hercules/resource/resource_utilities.py` that do not download data
- [ ] If necessary (probably not) update docs


paulf81:
This PR labeled on hold and propose keeping this only for reference and implement incrementally via the process started in #246

PR #224, opened 2026-03-04
Multiunit thermal plants

### LINKED ISSUE #213: New thermal plant class that models down to the turbine level
Currently, thermal plants are modeled in Hercules as homogenous plants of the same turbines that are all operated the same way. This new class would enable the simulation of a thermal plant down to the turbine level. This would include

- The ability to simulate individual turbines (collect and simulate more than one of the turbine models in a plant)
- The ability to create a plant composed of multiple types of turbines 
- The ability to dispatch these turbines differently according to their type and dynamics

misi9170:
This pull request adds functionality for having multiple thermal generating units under a single thermal power plant. This could be several independent units (e.g. two open-cycle gas turbines) or _dependent_ units, such as the gas turbine and the steam turbine in a combined cycle gas plant. The latter is not currently included, although @jfrederik-nrel is working on it.

As part of this PR, @jfrederik-nrel has also added a generic steam turbine unit. Final validation and documentation of this will be added before we mark this PR as ready for review.
EDIT: as per discussion below, the `SteamTurbine` class was removed from this PR and moved to #232 instead.


hercules/plant_components/thermal_plant.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.units = []
+        for unit, unit_name in zip(h_dict[component_name]["units"], self.unit_names):
+            h_dict_thermal = h_dict[component_name]
+            h_dict_thermal["dt"] = h_dict["dt"]
+            h_dict_thermal["starttime"] = h_dict["starttime"]
+            h_dict_thermal["endtime"] = h_dict["endtime"]
+            h_dict_thermal["verbose"] = h_dict["verbose"]
+            unit_type = h_dict['thermal_power_plant']['OCGT1']['component_type']
+            unit_class = hp.COMPONENT_REGISTRY[unit_type]  # Validate that the unit type is in the registry
```
dzalkind: Test note.  Fix this with get_component_model.

misi9170:
To do:
- [x] Clean up docstrings
- [x] Add documentation (thermal plant, example)
- [x] Add tests
- [x] Finalize example

To consider:
- [x] ~Add @jfrederik-nrel 's combined cycle gas turbine model to this PR? Otherwise, we can open a separate PR for it.~ -> Decided to open a separate PR, #232.

misi9170:
@jfrederik-nrel , to keep things clear between this PR and #232, how about I remove the `SteamTurbine` model from this PR and add it back to your branch on #232. That way, I think I can pretty quickly clean this PR up so that it's ready fore review, which should make it easier for you to work on #232.

misi9170:
@dzalkind @jfrederik-nrel @genevievestarke this PR is now ready for review.

genevievestarke:
Move registry to a different place in Hercules?

hercules/plant_components/thermal_plant.py
```diff
@@ -0,0 +1,106 @@
+"""
+Multiunit thermal power plant.
+"""
+
+import copy
+
+import hercules.hybrid_plant as hp
```
misi9170: @genevievestarke , does [your comment about moving the registry](https://github.com/NatLabRockies/hercules/pull/224#issuecomment-4070022235) refer to changing this line so we don't have to import hybrid_plant here? I tried that for a while but kept running into circular import issues so gave up. It sounds like @paulf81 also has had this issue before. I do agree though that this seems weird, and it would be better if the registry was in its own file. Perhaps we create a separate issue for that?

genevievestarke: Yeah, @dzalkind mentioned that you were looking for a fix, but we didn't think there was one implemented yet. I'll make an issue!

examples/09_multiunit_thermal_plant/hercules_input.yaml
```diff
@@ -0,0 +1,62 @@
...
+starttime_utc: "2020-01-01T00:00:00Z"  # Midnight Jan 1, 2020 UTC
+endtime_utc: "2020-01-01T06:00:00Z"  # 6 hours later
+verbose: False
+log_every_n: 1
+
+plant:
+  interconnect_limit: 100000  # kW (100 MW)
+
+thermal_power_plant:
+  component_type: ThermalPlant
```
genevievestarke, line 21: Update to use three units where two use the same input and one uses it's own.

docs/examples/09_multiunit_thermal_plant.md
```diff
@@ -0,0 +1,27 @@
+# Example 09: Multi-unit Thermal Plant
```
genevievestarke, line 1: Update example docs

docs/thermal_plant.md
```diff
@@ -0,0 +1,94 @@
+# Thermal Plant
+
+The `ThermalPlant` class models generic single or multiunit thermal power plants. It expects to be assigned one or more thermal units, for example [`OpenCycleGasTurbine`s](open_cycle_gas_turbine.md). The individual units are established in the YAML configuration file,
+and may be repeats of the same type of units or heterogeneous units.
+
+Set `component_type: OpenCycleGasTurbine` in the component's YAML section. The section key is a user-chosen `component_name` (e.g. `open_cycle_gas_turbine`); see [Component Names, Types, and Categories](component_types.md) for details.
```
genevievestarke: Add a bit more transition to this paragraph.

hercules/plant_components/thermal_plant.py
genevievestarke: See if moving this to a new file is useful.

genevievestarke (commented):
Looks good, I just left some comments so I remember which edits I wanted. I'll be working on applying these today.

hercules/plant_components/thermal_plant.py
```diff
@@ -0,0 +1,109 @@
...
+                h_dict[component_name][unit_name] = copy.deepcopy(h_dict[component_name][unit])
+
+        # Remove the template from the component dict since it's now copied into each unit dict
+        for unit in generic_units:
+            if unit in h_dict[component_name]:
+                del h_dict[component_name][unit]
+
+        # Load component registry here to define units in thermal plant
+        # NOTE: this breaks a circular dependency issue
+        from hercules.component_registry import COMPONENT_REGISTRY
```
genevievestarke, line 49: @paulf81 Thoughts on moving the component registry to a new file and adding this import statement later in thermal_plant?

paulf81: Yes that works for me

genevievestarke (approved):

PR #226, opened 2026-03-05
Add initial coal plant model

genevievestarke:
This model adds a class to represent coal plant using `ThermalPlantBaseClass` as an initial starting point. This class defines defaults for coal plant operation based on the resource cited in the class.

- [x] Add class
- [x] Update docs
- [x] Add tests
- [x] Example (if needed)


paulf81 (approved):
Thank you!

PR #228, opened 2026-03-05
Add power playback component

### LINKED ISSUE #190: Power pass through option for all Hercules components
Create power pass through options for all Hercules components if only the power history is available (only for electricity-producing technologies)

- [ ] Solar
- [ ] Battery
- [ ] Natural Gas

paulf81:
This PR adds a new component `PowerPlayback` which simulates a generic generating component (so `component_category==generator`) by playing back pre-recorded SCADA data.  Assumes the SCADA input file contains a column `time_utc' and 'power' (in kW).

This PR also offered a chance to follow the instructions in `docs/adding_components.md` and can confirm these were accurate.

In addition to the new component, several tests are included, a new example and updated docs.


examples/inputs/07_generate_scada_power_playback.py
misi9170, line 1: I think there is already an 03 in the inputs? Also, I guess a more conceptual question---for scripts that generate inputs that are only used in a single example, should they just live within the examples themselves? That might not follow the current structure though.

hercules/plant_components/power_playback.py
```diff
@@ -0,0 +1,172 @@
+# PowerPlayback is based on WindFarmSCADAPower but models a generic power source
+# Note it is limited to playing back the prerecorded power,
+# there is no option to control.
```
misi9170, line 3: Should WindFarmSCADAPower be reworked to be a subclass of PowerPlayback? Or something?

docs/power_playback.md
```diff
@@ -0,0 +1,68 @@
+# Power Playback
+
+`PowerPlayback` plays back pre-recorded power data from a file, treating the recorded values as the output of a generic power-generating unit. There is no control interface — the power values are replayed exactly as stored.
+
+This component is useful for incorporating measured or externally modeled power output into a hybrid plant simulation without needing to model the underlying physics of the generator.
+
```
genevievestarke, line 6: Maybe we can add a note here like:
"The PowerPlayback model is intended for use for using the power output from a generator component in a Hercules simulation. This component can be used for all generators, but will only take in the total plant power from a wind farm (and not individual turbine powers). For individual turbine power granularity, please use the`WindFarmSCADAPower` class."

tests/power_playback_test.py
```diff
@@ -0,0 +1,46 @@
...
+    step_h_dict = {"step": 0}
+    step_h_dict["power_playback"] = {}
+
+    result = power_playback.step(step_h_dict)
+
+    # Verify outputs exist
+    assert "power" in result["power_playback"]
+
+    # Verify power
+    assert np.isclose(result["power_playback"]["power"], 1000.0)
```
genevievestarke, line 39: Is this for rounding purposes that it's not exact?

genevievestarke (commented):
Looks good, just a suggestion about clarity in the docs!

examples/inputs/07_generate_scada_power_playback.py
genevievestarke: This file has been renamed, but I made an issue on this larger question!

genevievestarke: #229

hercules/plant_components/power_playback.py
genevievestarke: It's a little unclear to me how much of the underlying code it would be helpful to have from PowerPlayback, but I think that would make sense in the long run to have WindFarmSCADAPower inherit from PowerPlayback. Documented in Issue #230

tests/power_playback_test.py
paulf81: Yes, but now that you point out, it's a copy paste and could been a more simple assert ==

docs/power_playback.md
paulf81: Done!

genevievestarke (approved):

PR #232, opened 2026-03-07
Add multi-unit combined cycle gas turbine component

jfrederik-nrel:
Work in progress. Still to do:

- [x] Merge latest version of `feature/mm-thermal` back in
- [x] Execute sanity check of the control logic
- [x] Clean up code, ruff
- [x] Complete the example, possibly merge with existing example
- [x] Write documentation
- [x] Add tests

No reviews necessary yet at this time.


jfrederik-nrel:
In the commit [f666194](https://github.com/NatLabRockies/hercules/pull/232/commits/f6661949e04a293a924ca782572ed65968d99de4), I have included logging of the plant efficiency in two different ways:
1. The efficiency of the combined cycle plant as a whole, similar to how the efficiency would be calculated for the single-unit CCGT model.
2. The efficiency as a sum of the efficiency of the gas and steam units individually. 

Calculating the efficiency both ways results in substantially different results, see figure below. This has two important reasons:
1. When calculating the efficiency of the CCGT as a single unit, it assumes that if the power is only at say 70%, it is running the whole unit at 70%. However, in the startup phase, when the gas turbine is running but the steam turbine is not yet running, it is actually running the gas turbine at 100% capacity while running the steam turbine at 0% capacity. I therefore believe that calculating the efficiency as the sum of the individual unit's efficiency is more accurate.
2. However, for this to work, we do need to change something for all thermal units. Right now, the code assumes the efficiency is equal to the lowest value in the table if power is zero. I would argue that in this case, the efficiency is either 0 or NaN, but definitely zero as long as fuel is still being burned. Due to this current definition at power=0, the figure below actually shows a higher-than-maximum (53% in this case) efficiency of the combined unit when the gas turbine is being turned on, where it really should be significantly lower than maximum.

With `power_output = self.efficiency_values[0]` if `power_output <= 0` in function `calculate_efficiency`: (legend of efficiency tabel: green is plant as a whole, red is as a sum of both units individually)
image

With `power_output = 0.0` if `power_output <= 0`:
image

I believe the bottom figure represents the most accurate representation of the actual efficiency, and therefore propose to change the efficiency to 0 for all thermal units when the power output is zero. However, there is probably still room for improvement on this method for calculating efficiency, so open to feedback (@paulf81 @genevievestarke @misi9170).

misi9170:
> In the commit [f666194](https://github.com/NatLabRockies/hercules/pull/232/commits/f6661949e04a293a924ca782572ed65968d99de4), I have included logging of the plant efficiency in two different ways:

I agree with the summed approach. I particularly like that you can see the effective increase in efficiency when the steam turbine comes online, which is a bit muted when using the plant efficiency as a whole (the efficiency only increases in that case because there is more power being generated):
image

Regarding efficiency at 0 power: I agree that any time power is zero (or less than zero?) and fuel consumption is greater than zero, efficiency should be zero. I wonder whether, as you suggest, we should also report that efficiency is NaN if the units are completely off (zero power, zero fuel consumption). That would seem reasonable to me: that state is a bit different to the case of fuel consumption with no power.

Possibly, this is simply a case of defining the efficiency interpolator to take a `fill_value` of `0` on the low end

jfrederik-nrel:
The only problem is that we currently do not define the fuel usage independently. Instead, the fuel usage is determined using the efficiency table and the power production, so it is a bit of a chicken or egg story in that sense.

I think the easiest solution would be to have the efficiency linked to the state of the thermal unit. If it's off, efficiency is NaN, if its starting up (burning fuel but not producing energy yet), it is 0, and finally, if it _is_ producing energy, we use the interpolation we use now.

misi9170:
Ah good point, yes, that makes sense!

jfrederik-nrel:
The latest commit does exactly that @misi9170. This is what the efficiency look like now:

image

I feel like this still isn't perfect, but it might be sufficient for now.

Two notes:
1. I'm now using np.nansum to plot the combined unit efficiency. This is needed as I want the efficiency to be a number if the gas turbine is producing power while the gas turbine is off. However, it has the downside that it also makes all NaN's zeros when both units are off. We might need to make that more sophisticated later.
2. I haven't tested the changes to `calculate_efficiency` in `thermal_component_base.py` in any of the other examples yet

jfrederik-nrel:
In the latest commit, I ended up choosing a middle way between both methods. It now uses some if-else logic to determine the efficiency based on the state of the gas and steam unit. I honestly don't love this solution, but believe it's currently the best way to model the efficiency given the limitations of the framework (see issue #234).

jfrederik-nrel:
@genevievestarke @dzalkind it's ready for review now. I would particularly recommend taking a look at the new tests I formulated as this was a first time for me.

dzalkind:
Very nice contribution, @jfrederik-nrel! 

Which PR should this follow? #224? Should #225 also be included?

This might help the comparison in the meantime: https://github.com/misi9170/hercules/compare/feature/mm-thermal...jfrederik-nrel:hercules:feature/mu_ccgt?expand=1

docs/steam_turbine.md
```diff
@@ -0,0 +1,155 @@
...
+| `efficiency_table` | Average plant efficiency | [2,3] |
+
+### Default Efficiency Table
+
+The default HHV net plant efficiency table is based on [2,3]:
+
+| Power Fraction | HHV Net Efficiency |
+|---------------|-------------------|
+| 1.00 | 0.35 (35%) |
+| 0.5o | 0.32 (32%) |
```
dzalkind: Possibly just a typo with 0.5o

docs/_toc.yml
```diff
@@ -31,6 +31,8 @@ parts:
     - file: electrolyzer
     - file: thermal_component_base
     - file: open_cycle_gas_turbine
+    - file: steam_turbine
+    - file: combined_cycle_plant
```
dzalkind, line 36: I'm not seeing this combined_cycle_plant.md file in docs

examples/07_open_cycle_gas_turbine/hercules_runscript.py
dzalkind, line 1: Did we mean to remove this file?

jfrederik-nrel: Yes, as it is now included as one of the options in `07_thermal_plants`.

examples/07_thermal_plants/hercules_runscript.py
dzalkind, line 1: Now I see they are combined in this new directory.

hercules/plant_components/combined_cycle_plant.py
```diff
@@ -0,0 +1,401 @@
...
+        if "steam_turbine" not in generic_units:
+            raise ValueError(
+                "For the combined cycle plant, one of the units must be a steam turbine."
+            )
+        if "open_cycle_gas_turbine" not in generic_units:
+            raise ValueError(
+                "For the combined cycle plant, one of the units must be an open cycle gas turbine."
+            )
+
+        if len(generic_units) != 2:
```
dzalkind, line 38: We may want to relax this constraint in some cases in the future, where multiple gas generators could be combined to feed a steam unit.  We can probably work around this for now.

Will the code break in this case? I'm not sure we should change it now, but it might be good to know in the future.

hercules/plant_components/combined_cycle_plant.py
```diff
@@ -0,0 +1,401 @@
...
+        self.efficiency = self.calculate_efficiency(self.power_output)
+
+        self.fuel_volume_rate = self.calculate_fuel_volume_rate(self.power_output)
+        self.fuel_mass_rate = (
+            self.fuel_volume_rate * self.units[self.gas_turbine_index].fuel_density
+        )
+
+        # Update h_dict with outputs
+        h_dict[self.component_name]["power"] = self.power_output
+        # h_dict[self.component_name]["state"] = self.state.value
```
dzalkind, line 224: Do you want the state or is it not needed because each unit has its own state?

hercules/plant_components/combined_cycle_plant.py
```diff
@@ -0,0 +1,401 @@
...
+
+        h_dict[self.component_name]["power"] = self.power_output
+
+        # TODO: we likely want to save off data for the individual units to the
+        # h_dict as well. Will need to figure out how to do that.
+
+        return h_dict
+
+    def control(self, power_setpoint):
+        """"""
```
dzalkind, line 249: Maybe denote that this is the control of the combined plant?

hercules/plant_components/combined_cycle_plant.py
```diff
@@ -0,0 +1,401 @@
...
+        if not isinstance(power_setpoint, (int, float)):
+            raise ValueError("power_setpoint must be a number")
+
+        # Set gas turbine power setpoint
+        self.units[self.gas_turbine_index].power_setpoint = self.gas_power_ratio * power_setpoint
+        self.units[self.steam_turbine_index].power_setpoint = (
+            1 - self.gas_power_ratio
+        ) * power_setpoint
+
+        # TODO: we probably want to add an actual controller for the gas turbine
```
dzalkind, line 261: It looks like you did this TODO?

dzalkind (approved):
Overall, this looks good to me.  Nice job!

I like the power set point input addition.  It might be nice to have that documented for future reference, along with making sure your new docs are included.

Otherwise, I left some small comments in the files; they are pretty minor and often for my understanding.

jfrederik-nrel:
@dzalkind it should follow #224. #225 is related, but a stand-alone PR I'd say. 

I merged #224 into this branch recently, so I don't expect any major merge conflicts.

Added the missing `combined_cycle_plant` doc.

hercules/plant_components/combined_cycle_plant.py
jfrederik-nrel: No I didn't. I meant that right now it uses the basic `_control` logic defined in the `thermal_component_base` class, as there is no `control` method defined in the `open_cycle_gas_turbine` class. What I meant here is that we might want to add that at some point, or else rename `_control` in `thermal_component_base` to `control`.

hercules/plant_components/combined_cycle_plant.py
jfrederik-nrel: Each unit has it's own state. We would need to define a new state machine for the combined unit as it has complicated combined states like "gas turbine on, steam turbine off". I refrained from doing that and just left it at each unit having its own state

hercules/plant_components/combined_cycle_plant.py
jfrederik-nrel: Yes, I was thinking about that too. What I think I will try to do on the short term (though perhaps not in this PR) is allow at least 2 (or perhaps more? From what I can find, 2 is currently the max) gas units powering a single steam unit. I don't _think_ that would break anything, but definitely not excluding the possibility that it causes complications that I'm currently not overseeing.

docs/combined_cycle_plant.md
```diff
@@ -0,0 +1,241 @@
...
+
+Similar to the `ThermalPlant` class, the `CombinedCyclePlant` class does not have many default parameters. Key attributes that must be provided in the YAML configuration file are the `OpenCycleGasTurbine` and `SteamTurbine` `units`, which is a list that is used to instantiate the individual thermal units that make up the plant, and `unit_names`, which is a list of unique names for each unit. The number of entries in `units` and `unit_names` must match.
+
+However, unlike the base `ThermalPlant` class, it is recommended that some parameters are defined outside the individual thermal units. Since the gas turbine does not have its own individual fuel source, instead using the rest heat from gas turbine, it is not possible to specify the efficiency and fuel consumption of the steam turbine in the same way as done for individual components that are not linked. As a result, these outputs are instead calculated for the plant as a whole, necessitating an efficiency table for the unit as a whole. Note that this table is only used when both the gas and steam turbine are running. If only the gas turbine is running, the `OpenCycleGasTurbine` efficiency table is used instead.
```
genevievestarke: @jfrederik-nrel in the second sentence, should it be steam turbine first?

jfrederik-nrel: The yaml file should contain a unit of OCGT and a unit of ST. The order doesn't matter, if I understand your question correctly.

examples/08_multi_unit_thermal_plants/input_files/hercules_input_mu-ccgt.yaml
```diff
@@ -0,0 +1,142 @@
+# Input YAML for hercules
+# Explicitly specify the parameters for demonstration purposes
+
+# Name
+name: example_10
+
+###
+# Describe this simulation setup
+description: Combined Cycle Gas Turbine (CCGT) Example
```
genevievestarke, line 9: Is the current implementation simulating just one OCGT+ST? Can you simulate multiples of these pairs? It seems like only one can be defined per `CombinedCyclePlant`. If you can assign multiples of these in the multi-unit plant, then maybe `CombinedCycleUnit` would be a better name?

genevievestarke (commented):
Really good work @jfrederik-nrel ! I just had a question about whether you can simulate multiple combined cycle units in the same plant or not? To create a whole plant, do you just scale the OCGT rating and the ST rating to account for all the units in the plant?

docs/combined_cycle_plant.md
genevievestarke: Ah, I'm specifically asking about the wording of the documentation: "Since the gas turbine does not have its own individual fuel source, instead using the rest heat from gas turbine, ..."
This sounds like one of the gas turbine references should be referring to the steam turbine?

examples/08_multi_unit_thermal_plants/input_files/hercules_input_mu-ccgt.yaml
jfrederik-nrel: Currently, you can only have one OCGT+ST, so yes, you would need to scale them to model a more complicated setup that for example has two steam turbines (which I do think is actually relatively common). I did consider making it generalizable, but this would severely complicate the control logic as there are many ways the different units could be linked. I therefore chose to keep it simple for now and stick to a single OCGT+ST setup. The nomenclature doesn't seem overly important to me personally, but if you want to change the name to `CombinedCycleUnit`, I won't fight you on it.

genevievestarke: Ok, since it's meant to represent the whole plant, I'll keep it as is for now!

PR #233, opened 2026-03-10
Fix number check

paulf81:
This PR fixes a small issue where `isinstance(x, (int, float))` doesn't match `np.float32` values (which is what `hercules_float_type` is), so these checks would incorrectly reject valid numeric inputs.  This PR simply corrects each occurence to also check for `hercules_float_type`


paulf81:
@genevievestarke all tests pass and co-pilot review is clean.  Calling this one ready for review (it's a small change)

genevievestarke (approved):
Looks good to me!

PR #236, opened 2026-03-19
Including fuel volume rates when starting up / shutting down

### LINKED ISSUE #234: Improve fuel consumption calculations of thermal units
I believe that the current framework used to calculate the efficiency and fuel consumption rate of thermal units (gas turbines, coal plants, etc) has some limitations that are especially pronounced in the combined cycle case (see #232). 

*Current framework*
In the current framework, the efficiency is provided through a lookup table that defines the efficiency as a function of load percentage. I assume this method was chosen since such tables/graphs are readily available online for different thermal power generators, enabling easy implementation. The fuel consumption is then calculated by dividing the power by the heating value (HHV) of the fuel times the efficiency.

*Limitations*
Although easy to implement, I believe this method is flawed, as it reverses the order of things as they happen in reality. The efficiency depends on the fuel consumption, not the other way around. As I see it, this method works reasonable well when the generator is on and producing any level of power between full load and minimum stable load. However, it doesn't accurately model the actual fuel consumption in start-up or stopping states. It assumes it just operates at or below its lowest efficiency point in the lookup table, still giving a relatively high efficiency and therefore relatively low fuel consumption. However, as far as I understand, especially the startup phase requires significant fuel consumption while still producing no to very limited power. In other words: the efficiency should be very low in this phase of the process. As a result, the current framework doesn't accurately model the (fuel) price of starting up and shutting down a thermal power plant.

*Possible solutions*
1. Reverse the order to determining the efficiency based on the fuel consumption. This would require a major overhaul and more data on the fuel consumption of different types of generators in different phases of operation. For that reason, this might not be feasible.
2. Implement a slightly more complicated efficiency / fuel consumption logic. We keep the current framework in place if the unit is on, but switch to different implementations when the turbine is revving up or down (and perhaps also when it's shut down). This would only require minor changes to the code, and only one or two additional input variables (something like a `startup_fuel_consumption_rate` and `shutdown_fuel_consumption_rate`). 

Perhaps this is too complicated / detailed for the current level of modeling we are trying to do, in which case, feel free to close the issue. But I would still like to hear your opinion on this matter. 

Tagging @genevievestarke @misi9170 @paulf81 and @dzalkind.

jfrederik-nrel:
This PR (partly) fixes issue #234. 

Work in progress. To do:

- [x] Properly test it: still needs to be added to `__init__`
- [ ] Implement in all thermal classes (not just `thermal_component_base`)
- [x] Include it in the examples
- [x] Update documentation
- [ ] Update tests

How I implemented it now is that users can provide the `startup_fuel_fraction` and `shutdown_fuel_fraction` as a fraction of the rated fuel consumption. If not provided, it is assumed to be 0 (startup and shutdown costs no fuel). This is probably the easiest way to implement it, but perhaps not the easiest value to obtain. Alternatively, we could directly provide the `startup`/`shutdown_fuel_volume_rate`, which might be easier to obtain. 

Would appreciate feedback on which method sounds best @dzalkind, @genevievestarke. Both input methods should be similarly easy to implement.


hercules/plant_components/thermal_component_base.py
```diff
@@ -124,6 +124,10 @@ def __init__(self, h_dict, component_name):
         self.min_up_time = component_dict["min_up_time"]  # s
         self.min_down_time = component_dict["min_down_time"]  # s
 
+        # Extract optional parameters for startup and shutdown fuel fractions
+        self.startup_fuel_fraction = component_dict.get("startup_fuel_fraction", 0)
```
genevievestarke: Is this the current behavior? Where the starting and stopping fuel consumption is zero?

hercules/plant_components/thermal_component_base.py
```diff
@@ -607,12 +611,28 @@ def calculate_fuel_volume_rate(self, power_output):
...
-        # fuel_volume_rate (m³/s) = power (W) / (efficiency * hhv (J/m³))
-        # Convert power from kW to W (multiply by 1000)
-        fuel_m3_per_s = (power_output * 1000.0) / (efficiency * self.hhv)
-
-        return fuel_m3_per_s
+        rated_fuel_consumption_rate = (self.rated_capacity * 1000.0) / (
+            self.hhv * self.calculate_efficiency(self.rated_capacity)
+        )  # m³/s at rated capacity
+
+        if self.state == self.STATES.ON:
```
genevievestarke: It seems like we could make the ON state the default and call out the starting and stopping behavior with the ifs?

hercules/plant_components/thermal_component_base.py
```diff
@@ -607,12 +611,28 @@ def calculate_fuel_volume_rate(self, power_output):
...
+
+            # Calculate fuel volume rate using HHV net efficiency
+            # fuel_volume_rate (m³/s) = power (W) / (efficiency * hhv (J/m³))
+            # Convert power from kW to W (multiply by 1000)
+            # Ensure fuel rate is at least the startup fuel fraction when on
+            return max(
+                (power_output * 1000.0) / (efficiency * self.hhv),
+                rated_fuel_consumption_rate * self.startup_fuel_fraction,
+            )
+        elif self.state == self.STATES.OFF:
```
genevievestarke: I think this is taken care of on line 611. We could combine this one and the earlier one, perhaps?

docs/thermal_component_base.md
```diff
@@ -85,6 +85,8 @@ All parameters below are defined in the Hercules input YAML file. The base class
 | `hhv` | J/m³ | Higher heating value of fuel |
 | `fuel_density` | kg/m³ | Fuel density for mass calculations |
 | `efficiency_table` | dict | Dictionary containing `power_fraction` and `efficiency` arrays (see below). Efficiency values must be HHV net plant efficiencies. |
+| `startup_fuel_fraction` | fraction (0-1) | Optional, fuel consumption during startup, as a fraction of rated fuel consumption. Defaults to 0 |
```
genevievestarke: Move this to the new optional parameters section in the docs.

genevievestarke (changes requested):
This looks good @jfrederik-nrel! I just had a few comments before it's ready to come in!

hercules/plant_components/thermal_component_base.py
jfrederik-nrel: Yes, the current behavior automatically sets fuel consumption to 0 when the power is <=0. So if `startup_fuel_consumption` is not defined, we get the same response as we had before this PR.

genevievestarke: It looks like the original behavior uses fuel when it is ramping up and down (not just the ON status), but not before the ramp up starts:
image

The power output is above zero once ramping starts, so it goes past the zero power return.

genevievestarke: Also, if you change the startup_fuel_fraction to 0.9, it changes the fuel behavior for the whole simulation: 
image

Should we have limits on the fuel fraction numbers?

jfrederik-nrel:
@genevievestarke I think my latest push addresses your previous comments:

1. I removed the `max` operation in `calculate_fuel_volume_rate` to get rid of the undesired behavior when `startup_fuel_rate` is set very high.
2. I redefined how the efficiency is calculated, reversing the order (calculate fuel rate first, base efficiency on this fuel rate and the current power)
3. Created a new function `interpolate_efficiency` that basically has the functionality that the old `calculate_efficiency` had, i.e., determine the efficiency by interpolating the `power_fraction`-`efficiency` table. This method is now only used when the turbine is on.

Note that it is now failing some tests, as I defined the efficiency as `NaN` if fuel consumption is zero to prevent divide-by-zero errors. As a result we will need to redefine some of the tests that assume efficiency is the lowest value in the table when power is zero. If we like this solution, I can easily implement that, but I wanted to make sure that we like this solution before I spent time on doing that.

genevievestarke (approved):
Hey @jfrederik-nrel, thank you for putting this together! I looked over it and I think it looks great. I like how the behavior looks now!
I think we're good to merge!

PR #240, opened 2026-03-27
Add Linear Generator Model

dzalkind:
This PR adds a model for a linear generator, characterized by it's relatively flat efficiency curve and fast ramp rates and cooldown times.

A few parameters of this model are still uncertain, and we should reach out for clarification on them, including

- Efficiency: the curve is flat, a single value is provided by the Mainspring data sheet.
- Hot/warm/cold startup times: these are my best guesses, but should also be verified
- Time to reach a warm, cold off state: these are currently hard coded in ThermalComponentBase.  I suggest we parameterize these and add the following states (`OFF_HOT`, `OFF_WARM`, `OFF_COLD`), but should discuss.

I suggest we proceed with estimate and update the parameters in a future PR.

I added a DEFAULTS dict in `LinearGenerator` to simplify the checking and testing of these values.  No need to do this for past components, but it could be helpful for others going forward.  

In the example, I'm modeling a single linear generator to demonstrate its features.  Typically, it will be used in an array, but that is a separate feature/example and we can work on more control approaches in Hycon.  I'm open to feedback on the example.

I'm not sure if we want to specifically call out Mainspring, but that is where a lot of our information has come from.

TODO:
- [x] Merge #224 when ready
- [x] Finalize parameters
- [x] (discuss) and implement `OFF_*` states
- [x] Revert nrel-pysam stuff I was tinkering with or fix it


genevievestarke (approved):
I like the new state plotting method @dzalkind, very slick! I think this is good to come in from my perspective!

PR #250, opened 2026-04-14
Update Hercules solar model to accept pysam_options dictionary

genevievestarke:
This PR allows the pysam PVWatts model in Hercules to accept a dictionary that can define the parameters in the `SystemDesign` portion of the model. A good reference for what you can set using this dictionary can be found here: https://h2integrate.readthedocs.io/en/stable/technology_models/pvwattsv8_solar_pv.html 

This is fully back compatible with current configurations, but it now allows the solar farm parameters to be set with more precision if needed.

## To Do:

- [x] Update Hercules docs
- [x] Add a test for this 

Tests to add: 

- That an error is thrown if losses, tilt, or system capacity are provided
- That the solar plant uses the defaults if pysam_options is not provided
- That the defaults are updated if pysam_options is provided

### DIFF
--- a/examples/03_wind_and_solar/hercules_input.yaml
+++ b/examples/03_wind_and_solar/hercules_input.yaml
@@ -35,10 +35,27 @@ solar_farm: # The name of component object 1
   lon: -105.1778
   elev: 1829
   system_capacity: 30000  # kW (30 MW)
-  tilt: 0  # degrees
-  losses: 0
+  pysam_options:
+    array_type: 3.0  # single axis backtracking
+    azimuth: 180.0
+    dc_ac_ratio: 1.0  # Force to 1.0
+    losses: 0
+    module_type: 0.0  # standard crystalline silicon (hardcoded)
+    tilt: 0  # degrees
   log_channels:
     - power
+
+# solar_farm: # The name of component object 1
+#   component_type: SolarPySAMPVWatts
+#   solar_input_filename: ../inputs/solar_input.ftr
+#   lat: 39.7442
+#   lon: -105.1778
+#   elev: 1829
+#   system_capacity: 30000  # kW (30 MW)
+#   tilt: 0  # degrees
+#   losses: 0
+#   log_channels:
+#     - power
 
   initial_conditions:
 
--- a/examples/03_wind_and_solar/hercules_input.yaml
+++ b/examples/03_wind_and_solar/hercules_input.yaml
@@ -35,27 +35,16 @@ solar_farm: # The name of component object 1
   lon: -105.1778
   elev: 1829
   system_capacity: 30000  # kW (30 MW)
+  tilt: 0  # degrees
+  losses: 0
   pysam_options:
-    array_type: 3.0  # single axis backtracking
-    azimuth: 180.0
-    dc_ac_ratio: 1.0  # Force to 1.0
-    losses: 0
-    module_type: 0.0  # standard crystalline silicon (hardcoded)
-    tilt: 0  # degrees
+    SystemDesign:
+      array_type: 3.0  # single axis backtracking
+      azimuth: 180.0
+      dc_ac_ratio: 1.0  # Force to 1.0
+      module_type: 0.0  # standard crystalline silicon (hardcoded)
   log_channels:
     - power
-
-# solar_farm: # The name of component object 1
-#   component_type: SolarPySAMPVWatts
-#   solar_input_filename: ../inputs/solar_input.ftr
-#   lat: 39.7442
-#   lon: -105.1778
-#   elev: 1829
-#   system_capacity: 30000  # kW (30 MW)
-#   tilt: 0  # degrees
-#   losses: 0
-#   log_channels:
-#     - power
 
   initial_conditions:
 
--- a/examples/03_wind_and_solar/hercules_input.yaml
+++ b/examples/03_wind_and_solar/hercules_input.yaml
@@ -42,7 +42,7 @@ solar_farm: # The name of component object 1
       array_type: 3.0  # single axis backtracking
       azimuth: 180.0
       dc_ac_ratio: 1.0  # Force to 1.0
-      module_type: 0.0  # standard crystalline silicon (hardcoded)
+      module_type: 0.0  # standard crystalline silicon
   log_channels:
     - power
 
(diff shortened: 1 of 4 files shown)


hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -43,20 +43,38 @@ def _setup_model_parameters(self, h_dict):
...
-                    "tilt": h_dict[self.component_name]["tilt"],
+        if h_dict[self.component_name].get("pysam_options") is not None:
+            print(
+                "PySAM model options provided in input are being used to define the PVWatts system."
+            )
+            if "losses" in h_dict[self.component_name] or "tilt" in h_dict[self.component_name]:
+                print(
+                    "Warning: 'losses' and 'tilt' parameters in the input will be ignored "
+                    "since PySAM model options are provided."
+                )
```
paulf81: For consistency should use self.logger.info in place of print

paulf81: And on the warning statement self.logger.warning()

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -43,20 +43,38 @@ def _setup_model_parameters(self, h_dict):
...
+            sys_design = {
+                "ModelParams": {"SystemDesign": h_dict[self.component_name]["pysam_options"]},
+            }
+
```
paulf81: Do we need a check that sys_design contains some minimum number of defined parameters?  We could also supply defaults through an |

examples/03_wind_and_solar/hercules_input.yaml
```diff
@@ -35,11 +35,28 @@ solar_farm: # The name of component object 1
...
+    array_type: 3.0  # single axis backtracking
+    azimuth: 180.0
+    dc_ac_ratio: 1.0  # Force to 1.0
+    losses: 0
+    module_type: 0.0  # standard crystalline silicon (hardcoded)
+    tilt: 0  # degrees
   log_channels:
     - power
 
+# solar_farm: # The name of component object 1
```
paulf81: Let's remove commented code before final merge

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -43,20 +43,38 @@ def _setup_model_parameters(self, h_dict):
...
+            # Ensure system capacity is set from upper level of input
+            # TODO: Should losses and tilt also be forced to be set from
```
paulf81: What if we say, everything that is in effect implemented through sys_design must come in through pysam_options to make things less confusing

paulf81:
Thank you @genevievestarke so much for jumping on this!!  Had a few comments for your consideration and happy to help with any pieces like logging etc.,

hercules/plant_components/solar_pysam_pvwatts.py
genevievestarke: Hmm, I was trying to differentiate between the pysam_options definition and the Hercules defaults. We could have defaults for some parameters in pysam_options and not others, but that might obscure what you can actually set. Alternatively, we could require the users to set all of the necessary parameters every time. Thoughts?

paulf81: My idea was if a parameter is only used by Hercules as a pass-through to pvwatts, it goes into the pysam_options sub_dict, otherwise if it's used at least one other place within Hercules can be higher level and copied in

hercules/plant_components/solar_pysam_pvwatts.py
genevievestarke: Done! Let me know what you think!

examples/03_wind_and_solar/hercules_input.yaml
genevievestarke: Done!

hercules/plant_components/solar_pysam_pvwatts.py
genevievestarke: Done! One is updated for a logger.info and the other was changed to a ValueError

paulf81: looks great, thanks!

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -43,18 +43,45 @@ def _setup_model_parameters(self, h_dict):
...
+            if len(common_keys) > 0:
+                raise ValueError(
+                    f"Error: The following parameters are provided in both the top-level input\
+                        and the PySAM options: {common_keys}. Please remove these parameters\
+                        from the PySAM options."
+                )
+
+        model_dict = (
+            top_level_dict
+            | hercules_defaults
```
paulf81: I feel like hercules_defaults should be first, so always over-written first.  I get they don't overlap, but if that ever changes this feels like the wrong behavior

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -43,18 +43,45 @@ def _setup_model_parameters(self, h_dict):
...
+            "system_capacity": system_capacity,
+        }
+        top_level_set = set(top_level_dict.keys())
+
+        # These values are the Hercules defaults for the PVWatts model and will be used if
+        #   not provided in the PySAM options in the input.
+        hercules_defaults = {
+            "array_type": 3.0,  # single axis backtracking
+            "azimuth": 180.0,
+            "dc_ac_ratio": 1.0,  # Force to 1.0
```
paulf81: This is now a default and not a force right?  I think I had set it to force not knowing what it did at the time, but now user is allowed to change right?

hercules/plant_components/solar_pysam_pvwatts.py
paulf81: yes, that looks really good!  I had a few more very small comments but all in all great!

hercules/plant_components/solar_pysam_pvwatts.py
paulf81: fixed

hercules/plant_components/solar_pysam_pvwatts.py
paulf81: thanks for fixing!

paulf81 (approved):
Looks great to me!

PR #254, opened 2026-04-29
Feature/substep pvwatts

paulf81:
The potential solar power is computed in advance in Hercules.  #249 corrected a bit how this was computed by noting that when upsampling 15 min / 1 hour solar input data to the hercules 1/10S, it was most appropriate to interpolate from the midpoint of the time period, since the timestamp marks the beginning of the period.

The primary objective of this PR is to delay the upsampling of the solar data from before calling pysam/pvwatts to after.  This is because recent experience shows that the over-running of pvwatts on identical inputs which was cheap previously can be expensive if bifaciality is applied.   In adding this feature, the opportunity is taken to correct a slight issue from #249 that pvwatts is also assuming timestamps are start of period and shifting to the midpoint for solar calculations.  This doesn't mean the previous code has to change but some of the comments were slightly confused and are clarified in detail here.  

A new option for the solar module is included `use_native_solar_dt`, which defaults to `True` and uses the new method of running pvwatts on the underlying solar time step before upsampling to the hercules time stamp.  If `use_native_solar_dt` is false, or the hercules time step is not smaller than the solar data time step, then the behavior should be equivalent to the beahavior from before.  A new test module is added to confirm that while the new method saves a lot of computation time, it's output is very similar to the previous behavior and should probably be always used (since calling pvwatts multiple times with identical inputs and only small changes in solar position should not impact results too much)

The documentation is also updated.


hercules/plant_components/solar_pysam_base.py
```diff
@@ -120,10 +120,50 @@ def _load_solar_data(self, h_dict):
         # Set starttime_utc (zero_time_utc is redundant since time=0 corresponds to starttime_utc)
         self.starttime_utc = starttime_utc
 
-        # Interpolate df_solar on to the time steps
-        time_steps_all = np.arange(self.starttime, self.endtime, self.dt, dtype=hercules_float_type)
+        # Determine the dt implied by the weather file (after sorting to be safe)
+        df_solar = df_solar.sort_values("time").reset_index(drop=True)
```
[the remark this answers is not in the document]
paulf81: Good catch, seems like the edgiest edge case but taking the suggestion

hercules/plant_components/solar_pysam_base.py
```diff
@@ -120,10 +120,50 @@ def _load_solar_data(self, h_dict):
...
+        # ``averaged_to_instantaneous`` upsample never has to extrapolate.
+        # In the fallback (compute_dt == dt) the compute grid equals the
+        # Hercules grid so downstream array lengths match step indexing
+        # exactly, preserving the pre-existing behaviour.
+        if self.use_native_solar_dt and self.dt_solar > self.dt:
+            self._compute_dt = self.dt_solar
+            n_compute = max(int(np.ceil((self.endtime - self.starttime) / self._compute_dt)) + 1, 2)
+            self._compute_time_steps = self.starttime + np.arange(
+                n_compute, dtype=hercules_float_type
+            ) * hercules_float_type(self._compute_dt)
```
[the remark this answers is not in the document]
paulf81: I think I've now accounted for this but you should recheck

tests/solar_pysam_pvwatts_resource_dt_test.py
```diff
@@ -0,0 +1,200 @@
...
+    PVWatts-at-Hercules-dt path. The only residual difference comes from
+    PVWatts' internal sun-position half-step shift moving from
+    ``dt_hercules / 2`` to ``dt_compute / 2``, which is small for short
+    sim windows away from sunrise/sunset.
+    """
+    starttime_utc = pd.to_datetime("2024-06-24T17:00:00Z")
+    endtime_utc = pd.to_datetime("2024-06-24T17:10:00Z")
+    dt = 1.0
+    dt_native = 60.0
+
```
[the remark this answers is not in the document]
paulf81: I think this is now accounted for but you should recheck

hercules/plant_components/solar_pysam_base.py
```diff
@@ -120,10 +120,58 @@ def _load_solar_data(self, h_dict):
...
+        # exactly, preserving the pre-existing behaviour.
+        if self.use_native_solar_dt and self.dt_solar > self.dt:
+            self._compute_dt = self.dt_solar
+            native_time = df_solar["time"].to_numpy(dtype=hercules_float_type)
+            i_start = max(np.searchsorted(native_time, self.starttime, side="right") - 1, 0)
+            i_end = min(
+                np.searchsorted(native_time, self.endtime, side="left"), len(native_time) - 1
+            )
+            self._compute_time_steps = native_time[i_start : i_end + 1]
+            interpolation_method = "instantaneous_to_instantaneous"
```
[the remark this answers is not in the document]
paulf81: Fixed

hercules/plant_components/solar_pysam_pvwatts.py
```diff
@@ -119,23 +119,35 @@ def _precompute_power_array(self):
...
-        )
+        # Pull all outputs at the compute-grid resolution, with W -> kW for the
+        # power channels. AC (post-inverter) and DC (pre-inverter) are
+        # "available" power: the model output before any Hercules control
+        # curtailment.
+        compute_outputs = {
+            "ac_power_available": np.array(self.system_model.Outputs.ac, dtype=hercules_float_type)
+            / 1000.0,
+            "dc_power_available": np.array(self.system_model.Outputs.dc, dtype=hercules_float_type)
+            / 1000.0,
```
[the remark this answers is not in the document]
paulf81: taken

docs/solar_pv.md
```diff
@@ -20,6 +20,60 @@ The solar component requires a weather time-series file. Supported formats are C
...
+and its power and diagnostic outputs are upsampled to the Hercules grid.
+This avoids re-running PySAM tens of thousands of times against
+essentially-identical interpolated weather, which becomes the dominant cost
+once bifacial modeling is enabled.
+
+The behavior is controlled by an opt-out flag:
+
+```yaml
+solar_farm:
+  use_native_solar_dt: true   # default
```
genevievestarke:
```suggestion
  use_native_solar_dt: false   # defaults to true, opt out by using false
```

docs/solar_pv.md
```diff
@@ -20,6 +20,60 @@ The solar component requires a weather time-series file. Supported formats are C
...
+
+- The feature only activates when the file's native dt is strictly greater
+  than `dt`. When the native dt equals `dt` (or the flag is set to `false`),
+  Hercules falls back to the existing path that runs PySAM once per Hercules
+  step.
+- Native dt is auto-detected from the loaded weather file (the difference
+  between consecutive sorted timestamps).
+- The PySAM outputs are upsampled to Hercules dt using
+  `"averaged_to_instantaneous"` (the same midpoint-corrected interpolation
+  introduced in PR #249). This is the single boundary crossing from the
```
genevievestarke:
```suggestion
  introduced in PR \#249). This is the single boundary crossing from the
```

I don't think this will link to the PR in the docs, so adjusting to just render the pound sign

docs/solar_pv.md
```diff
@@ -20,6 +20,60 @@ The solar component requires a weather time-series file. Supported formats are C
...
+applies is internal sun-position math, which it computes at the midpoint of
+each input time step (see the SAM help page on
+[Time and Sun Position](https://samrepo.nlr.gov/help/weather_time_convention.html)
+and the
+[PVWatts Version 5 Manual](https://docs.nrel.gov/docs/fy14osti/62641.pdf)
+§"Sun Position"). Because PVWatts does not shift the *values* of the
+irradiance/temperature inputs, performing the single
+"averaged → instantaneous" boundary crossing on the PVWatts *outputs* (at
+Hercules dt) is numerically equivalent to performing it on the *inputs* (at
+Hercules dt and then running PVWatts at every Hercules step) in the
```
genevievestarke:
```suggestion
Hercules dt) and then running PVWatts at every Hercules step in the
```

docs/solar_pv.md
```diff
@@ -20,6 +20,60 @@ The solar component requires a weather time-series file. Supported formats are C
...
+linear-PVWatts limit. The new path also slightly cleans up a labeling
```
genevievestarke: I don't think we need this extra explanation since this change has only been implemented for a week or two.

genevievestarke (commented):
Ok, I don't have a ton of thoughts on the implementation yet, but I think `use_resource_solar_dt` is more clear than `native`. It's not clear here what native is referring to (in my opinion), and resource seems more precise.

docs/solar_pv.md
paulf81: done

docs/solar_pv.md
paulf81: done

docs/solar_pv.md
paulf81: Removed

genevievestarke (approved):
Looks good!

PR #255, opened 2026-04-30
Add group tag

paulf81:
This PR does two things that shouldn't change Hercules behavior

1) Adds a new YAML key: `component_group` to the list of `component_name`, `component_type` and `component_category`.  Hercules itself won't use this group but it can be helpful to herc_analysis to flag that say several solar components are actually blocks of one solar farm.  If no group is specified then `component_group` = `component_name`.  Docs are updated and tests of new behavior included.

2) Since this change is done to benefit herc_analysis, a new doc page describing herc_analysis is included to mention and link to it.  This can then explain the purpose of the new `component_group` key.

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -53,3 +53,6 @@ parts:
     - file: examples/06_wind_and_hydrogen
     - file: examples/07_open_cycle_gas_turbine
     - file: examples/09_multiunit_thermal_plant
+  - caption: Post-Processing
+    chapters:
+    - file: herc_analysis
--- a/docs/adding_components.md
+++ b/docs/adding_components.md
@@ -83,7 +83,7 @@ class MyComponent(ComponentBase):
 | Requirement | Description |
 |---|---|
 | `component_category` | Must be `"generator"`, `"storage"`, or `"load"`. Generators contribute to `locally_generated_power`. |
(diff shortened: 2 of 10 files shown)


misi9170:
#224 introduced "grouped" generating units by actually making them subsets of a component. This seems to be rather similar functionality. Could the multiunit plants introduced in #224 instead be flattened and just share a `component_group` tag, or would that change what is possible with a multiunit plant? Alternatively, should we simply create a generic multiunit component type to replace your `component_group`, and assign all grouped components to the same higher-level multiunit component (something we talked about when implementing #224)? This would mean that, say, we could have a single top-level solar component with multiple different lower level solar units, in a hierarchy.

paulf81:
> #224 introduced "grouped" generating units by actually making them subsets of a component. This seems to be rather similar functionality. Could the multiunit plants introduced in #224 instead be flattened and just share a `component_group` tag, or would that change what is possible with a multiunit plant? Alternatively, should we simply create a generic multiunit component type to replace your `component_group`, and assign all grouped components to the same higher-level multiunit component (something we talked about when implementing #224)? This would mean that, say, we could have a single top-level solar component with multiple different lower level solar units, in a hierarchy.

This is a good point.  Noting you and I discussed this some IRL and going to put here I'm going to put this PR on hold, maybe permanently, in favor of a longer term plan something more like you propose here.

PR #261, opened 2026-06-04
Update GitHub workflow to enable ruff format testing for all python versions

genevievestarke:
Previously, ruff formatting was only running for python v3.13, which was leading to some confusing testing errors. This PR enables ruff formatting to be run on all python versions.


misi9170 (approved):
Good for me, but I'd go ahead and just delete the line rather than commenting it out (up to you, though :) )

paulf81 (approved):

